PR #70, opened 2015-01-26
AC-82

ehaligowska:


KamilKalfas:
This PR should contain only AC-82 commits.

PR #105, opened 2016-02-12
AC -141 and AC-143 Fixed

AvijitGhosh82:
AC-141 Reduced icon size
AC-143 Fixed the style values for new appcompat. Added cards to match web ui. Colors fixed.


AvijitGhosh82:
@rkorytkowski Is this all right?

openmrs-client/src/main/java/org/openmrs/mobile/activities/DashboardActivity.java
```diff
@@ -57,6 +59,11 @@ protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_dashboard);
         FontsUtil.setFont((ViewGroup) findViewById(android.R.id.content));
+        getSupportActionBar().setDisplayShowHomeEnabled(true);
+        //getSupportActionBar().setLogo(R.drawable.ic_openmrs);
```
rkorytkowski: Please do not leave commented out code. If you think it should be deleted, then delete it.

openmrs-client/src/main/res/diiflayouts/activity_dashboard.txt
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<!--<?xml version="1.0" encoding="utf-8"?>-->
```
rkorytkowski: Why did you comment this out?

AvijitGhosh82: These were layout files for large and xlarge type devices. Since that is not the correct norm (now) for creating tablet UI, I have moved them to a text file, and will work on creating a tablet UI once we have a dynamic fragment system ready , instead of activities ( values -820 dp). This is a text file and not an xml file, which means it can be treated as a note/todo for future redesign.

rkorytkowski: So it's not used at the moment, right? Could you please make it clear by adding a comment at the beginning of the file with some explanation?

AvijitGhosh82: Added. Please check commit.

AvijitGhosh82:
@rkorytkowski  Is this all right?

rkorytkowski:
The branch needs to be updated to the latest version of master before it can be merged.

openmrs-client/src/main/AndroidManifest.xml
```diff
@@ -25,6 +25,8 @@
     <uses-permission android:name="android.permission.READ_PHONE_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
 
+    <!--android:theme="@style/Theme.Openmrs"-->
```
rkorytkowski: When updating the branch, please remove this comment as well.

AvijitGhosh82:
@rkorytkowski Level with master done. Also removed code comment.

PR #108, opened 2016-02-16
Fixed AC 139 - Added instruction dialog when user forgets password

AvijitGhosh82:


AvijitGhosh82:
Cleaned and sent new PR #109

PR #110, opened 2016-02-18
Fix AC-146 and AC-147

AvijitGhosh82:
Patient list loading crashed if the birthday was null (AC-147)
The app did not connect to server if user did not explicitly mention http or https (AC-146)
Both fixed.

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/LoginActivity.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/LoginActivity.java
@@ -46,9 +46,6 @@ import org.openmrs.mobile.utilities.FontsUtil;
 import org.openmrs.mobile.utilities.ImageUtils;
 import org.openmrs.mobile.utilities.ToastUtil;
 import org.openmrs.mobile.utilities.URLValidator;
-
-import java.net.MalformedURLException;
-import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/activities/LoginActivity.java
```diff
@@ -285,6 +289,15 @@ public void setUrl(String url) {
         }
     }
 
+    public String ensure_has_protocol(String a_url)
```
rkorytkowski: Please follow java conventions when naming methods. Do not use _.

openmrs-client/src/main/java/org/openmrs/mobile/net/AuthorizationManager.java
```diff
@@ -33,9 +38,17 @@ public void login(LoginListener listener) {
         String url = listener.getServerURL() + LOGIN_END_URL;
         mLogger.i(SENDING_REQUEST + url);
 
+        Response.ErrorListener errorListener = new Response.ErrorListener() {
+
+            @Override
+            public void onErrorResponse(VolleyError error) {
+                error.printStackTrace();
```
rkorytkowski: Handle it properly. Use http://developer.android.com/reference/android/util/Log.html

What error message is displayed to the user?

AvijitGhosh82: This particular code is for fixing AC-137 (Work in progress) and unrelated to the NPE. I've added the log commit though, but I'll examine the actual log later as I get time.

AvijitGhosh82:
Ping @rkorytkowski

rkorytkowski:
Could you please try sending one PR per issue (especially do not include any code related to other issues without mentioning the issue id in the commit message)? If it is a quick fix, you could also push without a PR and ask for a post commit review by adding the commit URL in the comment on the issue.

PR #114, opened 2016-02-23
Optimize Imports

AvijitGhosh82:
Rearranged imports and removed unused if any


AvijitGhosh82:
@rkorytkowski commit e1eee1b is where the changes are, I don't know why the other commits are showing up here, given that I've synced upstream before optimizing.

PR #120, opened 2016-02-24
Fix AC-148 : Assign proper resource id to view at runtime to prevent crashes.

AvijitGhosh82:


odkcollect/src/main/java/org/odk/collect/android/views/HierarchyElementView.java
```diff
@@ -68,6 +70,26 @@ public HierarchyElementView(Context context, HierarchyElement it) {
 
     }
 
+    public final AtomicInteger sNextGeneratedId = new AtomicInteger(1);
```
rkorytkowski: Please declare class fields at the top of the file. I see no reason to expose it as public here.

odkcollect/src/main/java/org/odk/collect/android/views/HierarchyElementView.java
```diff
@@ -68,6 +70,26 @@ public HierarchyElementView(Context context, HierarchyElement it) {
...
+
+	/**
+	 * Generate a value suitable for use in {@link #setId(int)}.
+	 * This value will not collide with ID values generated at build time by aapt for R.id.
+	 *
+	 * @return a generated ID value
+	 */
+	public int generateNewViewId() {
```
rkorytkowski: Could you please check what's up with formatting here? Your code seems to be differently aligned than the rest of class.

AvijitGhosh82: Edited the formatting to match the rest of the file. This code was directly lifted from View.java of Android-17 where this bug was fixed in the source and so their formatting is a bit different. This function is kind of a backport for older apis and Google recommends to use it now.

PR #122, opened 2016-02-25
Added http url addressing for query parameter

pulkit4tech:
This is an implementation of bug filed over at https://issues.openmrs.org/browse/AC-145

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/net/helpers/FindPatientsHelper.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/net/helpers/FindPatientsHelper.java
@@ -20,6 +20,7 @@ import org.openmrs.mobile.activities.PatientDashboardActivity;
 import org.openmrs.mobile.listeners.findPatients.FindPatientListener;
 import org.openmrs.mobile.listeners.findPatients.FullPatientDataListener;
 import org.openmrs.mobile.listeners.findPatients.LastViewedPatientListener;
+import java.net.URI;
 
 public final class FindPatientsHelper {
 
@@ -27,6 +28,12 @@ public final class FindPatientsHelper {
     }
 
     public static FindPatientListener createFindPatientListener(String lastQuery, int searchId, FindPatientsSearchActivity caller) {
+        try {
+            URI uri = new URI(null,null,null,lastQuery);
+            lastQuery = uri.toString().substring(1);
+        }catch (Exception e){
+            e.printStackTrace();
+        }
         return new FindPatientListener(lastQuery, searchId, caller);
     }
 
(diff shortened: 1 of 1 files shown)


AvijitGhosh82:
Have you verified if this adds HTTP to the URL if not present? I tried to use this in a previous PR, but apparently calling URL() does not add protocol headers. I had to use regex matching to fix it. 

Also, is URL(null,null,null, String) required? URL (String) should have worked I believe.

pulkit4tech:
Yes we can add Http and protocol header using URI and URIBuilder class.But i saw that url was correct ,except the query parameter which is adding like a string containing whitespace which is giving 505 error while calling.So to fix it I just tried to format just parameters as rest of url was fine.
I tried using just URI(String) but it didn't worked as first parameter is for scheme.

openmrs-client/src/main/java/org/openmrs/mobile/net/helpers/FindPatientsHelper.java
```diff
@@ -27,6 +28,12 @@ private FindPatientsHelper() {
     }
 
     public static FindPatientListener createFindPatientListener(String lastQuery, int searchId, FindPatientsSearchActivity caller) {
+        try {
+            URI uri = new URI(null,null,null,lastQuery);
```
rkorytkowski: It may be cleaner to use http://developer.android.com/reference/java/net/URLEncoder.html

openmrs-client/src/main/java/org/openmrs/mobile/net/helpers/FindPatientsHelper.java
```diff
@@ -27,6 +28,12 @@ private FindPatientsHelper() {
...
+            lastQuery = uri.toString().substring(1);
+        }catch (Exception e){
+            e.printStackTrace();
```
rkorytkowski: Never eat exceptions. Report them to the user and log.

dkayiwa: Besides, exceptions can be poisonous for human consumption! :)

rkorytkowski: And infectious for others! :)

openmrs-client/src/main/java/org/openmrs/mobile/net/helpers/FindPatientsHelper.java
AvijitGhosh82: Or even https://github.com/openmrs/openmrs-contrib-android-client/blob/master/openmrs-client/src/main/java/org/openmrs/mobile/utilities/URLValidator.java 

It is far more extensive and developed in-house. See if this works? :)

openmrs-client/src/main/java/org/openmrs/mobile/net/helpers/FindPatientsHelper.java
dkayiwa: Lolll. :)

pulkit4tech: Thanks for review,I will be applying suggestion.

pulkit4tech:
Updated parameter encoding with URLEncoder as suggested and also added log for any error.

PR #123, opened 2016-02-26
Properly encoded query parameter in find patient request

joecheatham:
Addressed AC-133 by URL encoding the name query parameter in the "Find Patient" search.


rkorytkowski:
Possibly fixed already by AC-145.

joecheatham:
Appears to be fixed by AC-145.

PR #131, opened 2016-03-04
Fixed AC-154: Corrected the string format.

brainbreaker:
Previously message text in the dialog was: There are already active visit. Synchronize patient or close previous visit.
Now the warning/error in "ACBaseActivity.java" has been removed. We were not using the argument title declared in the function. Now the new message text in dialog box is:  There are already active visit for Bob Smith. Synchronize patient or close previous visit.


brainbreaker:
Thanks for reviewing Avijit. I have corrected the grammar. :)

AvijitGhosh82:
Merged. :)

PR #134, opened 2016-03-05
Fixed AC-138: We don't have to block UI by showing a progress dialog

brainbreaker:
A toast would be sufficient informing download has been started and ended. It doesn't blocks the UI and is more intuitive for the user.


AvijitGhosh82:
Good job! Works. Merging.

PR #136, opened 2016-03-06
Fixed AC-158:Too much margin on some activities, not so good looking

shivtej1505:


AvijitGhosh82:
The layout is fine as it is. The standard margin is recommended as default in the SDK docs. This change is not very important and so I'll not merge it in. Thanks for the PR though. :)

PR #140, opened 2016-03-15
Completed AC-160 (design and bugfix) and added Material icons for AC-159

AvijitGhosh82:


openmrs-client/src/main/java/org/openmrs/mobile/activities/fragments/FindPatientLastViewedFragment.java
```diff
@@ -99,9 +104,23 @@ public void onScroll(AbsListView absListView, int firstVisibleItem,
         });
 
         FontsUtil.setFont((ViewGroup) mFragmentLayout);
+        //registerForContextMenu(mPatientsListView);
         return mFragmentLayout;
     }
 
+   /* @Override
```
rkorytkowski: Could you please always delete commented out code? We can always get back the code from git history if needed. Otherwise it unnecessarily clutters the code and always makes me wonder why it was left behind.

AvijitGhosh82: This was probably done by the committer before me as I had pulled from his local branch before finishing it up. I will send in a patch with a small design change and the committed code removed very quickly.

openmrs-client/src/main/java/org/openmrs/mobile/adapters/PatientArrayAdapter.java
```diff
@@ -46,6 +56,64 @@
     private List<Patient> mItems;
     private int mResourceID;
 
+    private boolean isAllSelected = false;
+    private ActionMode mActionMode;
```
rkorytkowski: We should really stop using 'm' for prefixing class members. We have IDEs to highlight class members so using 'm' doesn't make sense and makes the fields harder to read. I don't say to fix it right now, but please do not copy the pattern going forward.

shivtej1505:
Thanks you @rkorytkowski for feedback, I'll remember these tips.
Just for clarification, this pull request was the first version,which I intentionally created only for review, I planned to cover all the suggestion in second version, but had exams and couldn't have time.

PR #143, opened 2016-03-18
AC-163: click event listeners should be work differently according to which fragment they are currently running

shivtej1505:
[issue link](https://issues.openmrs.org/browse/AC-163)


AvijitGhosh82:
Good job fixing the longpress listeners from your AC-160 implementation. Merged.

PR #148, opened 2016-03-22
AC-167 Upgraded ListView to RecyclerView

brainbreaker:
Upgraded List View to Recycler View in Last Viewed Patients Fragment, Downloaded Patients Fragment, Capture Vitals Activity and Find Patient Search Activity. Adapters which were upgraded are Patient Array Adapter and Patient Hierarchy Adapter. Patient Array Adapter renamed to Patient Recycler View Adapter.

### DIFF
--- a/openmrs-client/build.gradle
+++ b/openmrs-client/build.gradle
@@ -36,6 +36,7 @@ dependencies {
     compile 'com.google.guava:guava:18.0'
     compile 'com.android.support:appcompat-v7:23.1.1'
     compile 'com.android.support:gridlayout-v7:23.1.1'
+    compile 'com.android.support:recyclerview-v7:23.1.1'
     compile files('libs/commons-codec.jar')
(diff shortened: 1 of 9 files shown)


AvijitGhosh82:
Could you squash both your commits in one pull request?

brainbreaker:
@AvijitGhosh82 I have made two PRs and issues for a reason. IMO modifying all the adapters at once will not be a good idea, it will lead to change of lot of things in every activity, fragment and adapter . Modifying most of the code base in one single commit will not be good. *In case* we face some error in future, it would be easy to debug if we have different commits for different ListViews. I think, basic usage of git is because in case of any error we can revert to the last stable build. :)

AvijitGhosh82:
It is the same change applied to more classes. Please move your second commit to the earlier PR and let's keep commit history clean! I will be testing your PR locally anyway, so it's not an issue if there are more number of changes. I have merged both the issues as well.

PR #150, opened 2016-03-30
Fixed AC-169: 'Checking Visits' Progress Dialog does not dismiss

brainbreaker:
Fixed AC-169: 'Checking Visits' Progress Dialog does not dismiss. Instead of showing a Toast, we should dismiss the progress dialog.


AvijitGhosh82:
Works. Merging.

PR #151, opened 2016-03-30
 AC-170: Added animation to Recycler View Items when they appear

brainbreaker:
AC-170: Added Slide_in_left animation to Recycler View Items when they appear.


AvijitGhosh82:
Works. Merging in.

PR #153, opened 2016-04-01
Fixed AC-171

brainbreaker:
Fixed AC-171: Trimmed the url at `getText()` and made sure uppercase letters are converted to lower case. `ensureProtocol()` function moved to `URLValidator` class.

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/LoginActivity.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/LoginActivity.java
@@ -218,7 +218,7 @@ public class LoginActivity extends ACBaseActivity {
                 AuthorizationHelper.createBundle(
                         mUsername.getText().toString(),
                         mPassword.getText().toString(),
-                        mUrlTextView.getText().toString());
+                        mUrlTextView.getText().toString().trim());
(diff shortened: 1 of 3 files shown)


AvijitGhosh82:
Is this necessary? Doesn't the URL(String) function do this already? Did you find a case where this fixed a bug?

brainbreaker:
@AvijitGhosh82 

- Try any valid server URL and just give a new line at the end. Server error will appear. 
- Try Devtest04.openmrs.org:8080/openmrs. This will not work too, although the URL is correct if you go with the browser.
- If server URL is wrong, the ensure protocol function adds multiple `http://`. 
- We have the function to remove spaces in URLValidator class that removes trailing Whitespaces but unfortunately it doesn't works. It doesn't goes into the while loop with a new line in the end. In this URL
`'http://demo.openmrs.org/openmrs
                                                              
                                                              '`
the boolean `trimmedUrl.endsWith(SPACE)` returns false. I am still working to find the cause of this problem.

- This is why I have used `trim()`. It will remove any tailing or leading or in between whitespaces and works for every case.

AvijitGhosh82:
Fair enough. Merging in.

PR #154, opened 2016-04-04
Fixed AC-172: 

brainbreaker:
Fixed AC-172: Added null checks and updated correct variable values on activity pause.


AvijitGhosh82:
Merged.

PR #155, opened 2016-04-05
AC 173: Added Toolbars and Material designed Tabs

AvijitGhosh82:


AvijitGhosh82:
Merging in after Travis builds have passed.

PR #156, opened 2016-04-18
Fix build errors due to gradle version difference

AvijitGhosh82:
Added a flag to project properties to ignore gradle version difference. Also upgraded to latest gradle distribution.


AvijitGhosh82:
Merged.

PR #166, opened 2016-05-09
AC-177: Fixed inverted list of fragments causing force close

AvijitGhosh82:


AvijitGhosh82:
@rkorytkowski I think this is fixing our problem.

PR #168, opened 2016-05-25
Bugfixes before release

AvijitGhosh82:


AvijitGhosh82:
@rkorytkowski Is this all right for release? I have also found and sent you the faulty xml file..I've hidden capture vitals in this PR as you requested.

openmrs-client/src/main/AndroidManifest.xml
```diff
@@ -173,5 +173,15 @@
...
+        <activity
+            android:label="@string/action_register_patient"
+            android:name=".activities.RegisterPatientActivity"
+            android:theme="@style/AppTheme"
+            android:configChanges="keyboardHidden|orientation|screenSize">
+            <!-- Parent activity meta-data to support 4.0 and lower -->
+            <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
+                android:value=".activities.DashboardActivity" />
+        </activity>
```
rkorytkowski: It would be better to not add register patient activity for the coming release. Is it at least disabled and not visible for the user?

AvijitGhosh82: Yes, it is hidden. I've forked this to another branch and re-enabled it. :)

PR #175, opened 2016-06-15
AC-180: Use DAO instead of cache

AvijitGhosh82:
@rkorytkowski I've replaced the cache class to only use DAO everywhere. This should work theoretically, except that the JSON response returned in the PatientRegistrationService class is sometimes randomly null, and sometimes it has missing fields. Before merging, could you take a look at what is going on with the json response when the registration request is made? Thanks.


openmrs-client/src/main/java/org/openmrs/mobile/services/PatientRegisterService.java
```diff
@@ -44,39 +49,53 @@ public PatientRegisterService() {
...
+//                                            File path=getExternalFilesDir(null);
+//                                            File file=new File(path,"daojson2.txt");
+//
+//                                            try {
+//                                                Files.write(text,file, Charset.forName("UTF-8"));
+//                                            } catch (IOException e) {
+//                                                e.printStackTrace();
+//                                            }
+
+                                            Toast.makeText(PatientRegisterService.this, "Patient created with UUID "+ newPatient.getUuid()
```
AvijitGhosh82: @rkorytkowski The newPatient object returned is either null, or seems to have missing fields like name, address or identifier. I cannot seem to figure out why this is happening. I printed the object to a text file (using the commented lines above) and found this out. 

Please take a look, thanks.

openmrs-client/src/main/java/org/openmrs/mobile/services/PatientRegisterService.java
```diff
@@ -44,39 +49,53 @@ public PatientRegisterService() {
 
     @Override
     protected void onHandleIntent(Intent intent) {
-        final List<Patient> tbd=new ArrayList<>();
         if(isNetworkAvailable()) {
-            patientList=cache.getPatientList();
+            patientList=new PatientDAO().getAllPatients();
```
rkorytkowski: You should create getPatientsToSync() and use a sql query with `where synced = false`

openmrs-client/src/main/java/org/openmrs/mobile/services/PatientRegisterService.java
rkorytkowski: Can you paste the response here?

openmrs-client/src/main/java/org/openmrs/mobile/databases/tables/PatientTable.java
```diff
@@ -30,12 +29,13 @@
     private static final int INSERT_COLUMNS_COUNT = 18;
 
     @Override
-    public String crateTableDefinition() {
+    public String createTableDefinition() {
         return CREATE_TABLE + TABLE_NAME + "("
                 + Column.ID + PRIMARY_KEY
+                + Column.SYNCED + Column.Type.BOOLEAN
```
rkorytkowski: What happens if we want to upgrade the app? Is the table dropped and recreated?

rkorytkowski: See one possible approach on how to handle migrations here http://stackoverflow.com/a/998652

AvijitGhosh82: I've changed the database version number. This should cause the dropping and recreating to happen.

rkorytkowski: Dropping is not the best solution, but I guess it's okay for now. I'll create a separate issue to introduce db migrations. We'll address that in the future.

openmrs-client/src/main/java/org/openmrs/mobile/services/PatientRegisterService.java
AvijitGhosh82: Response for newPatient:

{"identifiers":[{"display":"OpenMRS ID \u003d
1007JH","links":[{"rel":"self","uri":"
http://devtest04.openmrs.org:8080/openmrs/ws/rest/v1/patient/5b3d06e4-ad33-426a-a63e-03f768b17699/identifier/3c9fa912-2aca-4e86-9049-75c20d8a8fd7"}],"uuid":"3c9fa912-2aca-4e86-9049-75c20d8a8fd7"}],"person":{"addresses":[],"attributes":[],"birthdate":"1961-06-16T07:30:00.000-1100","birthdateEstimated":false,"gender":"M","names":[]},"resourceVersion":"1.8","voided":false,"display":"1007JH
- Abc Def","links":[{"rel":"self","uri":"
http://devtest04.openmrs.org:8080/openmrs/ws/rest/v1/patient/5b3d06e4-ad33-426a-a63e-03f768b17699
"},{"rel":"full","uri":"
http://devtest04.openmrs.org:8080/openmrs/ws/rest/v1/patient/5b3d06e4-ad33-426a-a63e-03f768b17699?v\u003dfull
"}],"uuid":"5b3d06e4-ad33-426a-a63e-03f768b17699"}

As you can see, fields are missing in the response itself.

rkorytkowski: Please attach the request as well. "normal" is not enough :)

AvijitGhosh82: Request:

{"identifiers":[{"identifier":"1007P6","identifierType":"05a29f94-c0ed-11e2-94be-8c13b969e334","location":"7f65d926-57d6-4402-ae10-a5b3bcbf7986","links":[]}],"person":{"addresses":[{"address2":"xx"}],"attributes":[],"birthdate":"1961-06-17T00:00:00.000+05:30","birthdateEstimated":false,"gender":"M","names":[{"familyName":"yxz","givenName":"Xyz"}]},"display":"Xyz
yxz","links":[]}

This is for a different patient as I had not saved it the first time.
Interestingly this time the response is null.

PR #185, opened 2016-08-19
Fixed the loader in Patient Dashboard

AvijitGhosh82:
@rkorytkowski , I still could not emulate the error AC-189, I tried nearly 5 times. But I added a null check to the loader, just in case it might be the culprit. It did cause a force close once.


AvijitGhosh82:
@rkorytkowski I think the instance error should be fixed now.

PR #193, opened 2016-09-15
AC-196 Create ResourceSerializer for custom Gson serializing

AdamGrzybkowski:
@rkorytkowski @AvijitGhosh82 @tmarzeion 
please see issue [description](https://issues.openmrs.org/browse/AC-196)

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/api/PatientService.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/api/PatientService.java
@@ -24,8 +24,10 @@ import org.openmrs.mobile.api.promise.SimpleDeferredObject;
 import org.openmrs.mobile.api.promise.SimplePromise;
 import org.openmrs.mobile.application.OpenMRS;
 import org.openmrs.mobile.dao.PatientDAO;
+import org.openmrs.mobile.models.Location;
 import org.openmrs.mobile.models.retrofit.Encountercreate;
 import org.openmrs.mobile.models.retrofit.IdGenPatientIdentifiers;
+import org.openmrs.mobile.models.retrofit.IdentifierType;
 import org.openmrs.mobile.models.retrofit.Patient;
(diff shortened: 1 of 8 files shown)


AvijitGhosh82:
This is great! The current solution is to include the serializer and deserializer inside the models itself, which has the obvious drawback that those functions have to be called manually while persisting and retrieval of the objects. This is a far better solution! Have you confirmed if it works with ActiveAndroid properly?

AvijitGhosh82:
As an example, look at the code for Encountercreate. It would benefit from a serializer. There are several other classes, especially the ones used by ActiveAndroid.

AdamGrzybkowski:
I've tested it for syncing patients and it works :) 
Since you know the code a little bit more then me it would be great If you could implement this for encounter and test it. I will try to do that later anyway :)

AvijitGhosh82:
The Patients model was not using ActiveAndroid as it already had a mapper class written. Could you try implementing the Serializer to any of the classes that use ORM? 
I'll take a look once again though. :)

AdamGrzybkowski:
I've been looking at the code and the only place where I could show you the usage is createEncounter method in RestApi, since this only works with retrofit. Unfortunately I am not able to log into devtest server to test my changes(on int02 I cannot do anything with started visit) As soon as I will be able to login to devtest I will update this PR. 

PS. Are you planning to move from Volley to Retrofit in the whole project  ?

AvijitGhosh82:
Yes, we are. That was the initial plan, but the amount of refactoring meant that I could only do the new features in Retrofit as proof of concept and change the others later.

Ideally, we would like to remove the mapper classes as well, and only use Retrofit and ActiveAndroid.

@AdamGrzybkowski , why aren't you using Devtest04?

AdamGrzybkowski:
@AvijitGhosh82 I'm not sure but it looks like the forms are not working on demo.openmrs ?

PR #195, opened 2016-09-21
AC-195 show dialog with similar patients when registering new one

AdamGrzybkowski:
This PR contains commit from AC-196 PR because it requires those changes.  
I've some questions about the workflow:

1.In the current release when you register new user, you are redirect to the dashbord activity. Why not to the PatientDashbordActivity ? 
2. Why do we fetch patient in PatientDashbordActivity by uuid if we want to support offline mode? If we have more then one not synced patients we will always open the oldest one not synced. I quess this is a bug.

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/ACBaseActivity.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/ACBaseActivity.java
@@ -24,7 +24,6 @@ import android.support.v4.app.FragmentManager;
 import android.support.v4.app.FragmentTransaction;
 import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.SwitchCompat;
-import android.util.Base64;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.widget.CompoundButton;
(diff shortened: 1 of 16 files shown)


AvijitGhosh82:
@AdamGrzybkowski , the workflow you are mentioning does sound like a better idea! While creating the new patient registration page this slipped my mind. Could you add a commit to fetch patient by ID (and not uuid) and open patientdashboardactivity just after a user is registered? 

Thanks!

AdamGrzybkowski:
Sure ! :)

AdamGrzybkowski:
@AvijitGhosh82 I have updated PR, this is almost ready. There 2 things left:
1. First we need to merge AC-203.
2. Could you help me with the similar_patient_row layout? I had to set background of each element to white to look it ok. But this is kind of a hack. Also what do you think about this layout? The strings with bold style means that those values are matching new patient that user tries to register. 

screenshot_1474641332

AvijitGhosh82:
@AdamGrzybkowski , you might need to do a pull before the next commit, there's a merge conflict :)

AdamGrzybkowski:
@AvijitGhosh82 done :)

AvijitGhosh82:
@AdamGrzybkowski is this pull request ready for testing or you want to add anything else? :)

AdamGrzybkowski:
Ready for testing :)

AvijitGhosh82:
Adam, I was testing it locally using demo.openmrs.org server, and I get
"Unauthorized" as toast when I try to register a new patient.

AdamGrzybkowski:
@AvijitGhosh82  I am able to successfully register new patient on demo.openmrs.org

AvijitGhosh82:
Great, it worked for me just now as well. Extra brownie points for making the similar patient view clickable. Good job! Merging :)

PR #211, opened 2016-10-13
AC-196 fixed serializer to work with all resources

AdamGrzybkowski:
Sooo, the first serializer didn't work at all... We've noticed that with @tmarzeion when working on changing Visit model to match retrofit. 
I have modified it now to work with all classes that are Resources. 
Here is how it works:
when you post Resource object via retrofit this custom serializer checks all declared fields. If there is a Resource with uuid set it will serialize it to uuid only. When uuid is null it will serialize full object. 
When declared field is a collection, serializer iterates through it and does the same as above. It the collection contains objects that are not Resources,  it will be serializes with default gson serializer.

@rkorytkowski @AvijitGhosh82 @tmarzeion 
What do you think about that ? 
This is made to make posting object easier. Instead writing custom serializers for each objects e.g Patient/Visit we can use only this one. 
I have added some unit tests and also tested it manually by posting Patient and Visit.

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/utilities/ResourceSerializer.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/utilities/ResourceSerializer.java
@@ -1,17 +1,89 @@
 package org.openmrs.mobile.utilities;
 
+import android.support.annotation.NonNull;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
 import com.google.gson.JsonSerializationContext;
 import com.google.gson.JsonSerializer;
+import com.google.gson.annotations.Expose;
 
 import org.openmrs.mobile.models.retrofit.Resource;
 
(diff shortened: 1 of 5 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/utilities/ResourceSerializer.java
```diff
@@ -1,17 +1,89 @@
...
+                            }
+                        }
+                    } catch (IllegalAccessException e) {
+                        e.printStackTrace();
+                    }
+                } else {
+                    try {
+                        srcJson.add(field.getName(), myGson.toJsonTree(field.get(src)));
+                    } catch (IllegalAccessException e) {
+                        e.printStackTrace();
```
PawelGutkowski: No error handling?

AdamGrzybkowski: uuu thanks @PawelGutkowski, forgot about that. I'll fix it tomorrow ;)

tmarzeion:
@AdamGrzybkowski It works fine :) I'd merge this ASAP since it is required to finish AC-213

tmarzeion:
Thx @tmarzeion for merging :)

PR #214, opened 2016-10-19
AC-204 fixes and clean up after applying mvp

AdamGrzybkowski:
- removing any android libraries form presenters
- fixed crushing app when rotating the device
- fixed loading forms when changing server url in loginActivity


tmarzeion:
Looks OK, merging.

PR #227, opened 2016-12-12
AC-235 Hide downloaded patients

AdamGrzybkowski:
I couldn't really hide patients from the adapter so I filter patients before passing them to the adapter. There is problem with deleting the whole row. I had to set heights, padding etc. of the row manually to look ok, but then I was getting some inconsistency errors when scrolling.

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/lastviewedpatients/LastViewedPatientsPresenter.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/lastviewedpatients/LastViewedPatientsPresenter.java
@@ -18,11 +18,15 @@ import android.support.annotation.NonNull;
 
 import org.openmrs.mobile.api.RestApi;
 import org.openmrs.mobile.api.RestServiceBuilder;
+import org.openmrs.mobile.dao.PatientDAO;
 import org.openmrs.mobile.models.retrofit.Patient;
 import org.openmrs.mobile.models.retrofit.Results;
 import org.openmrs.mobile.utilities.ApplicationConstants;
 import org.openmrs.mobile.utilities.StringUtils;
 import org.openmrs.mobile.utilities.ToastUtil;
+
+import java.util.LinkedList;
+import java.util.List;
 
 import retrofit2.Call;
 import retrofit2.Callback;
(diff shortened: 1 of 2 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/activities/lastviewedpatients/LastViewedPatientsPresenter.java
```diff
@@ -85,6 +89,16 @@ public void onFailure(Call<Results<Patient>> call, Throwable t) {
 
     }
 
+    private List<Patient> filterNotDownloadedPatients(List<Patient> patients) {
+        List<Patient> newPatientList = new LinkedList<>();
+        for (Patient patient: patients){
+            if(!new PatientDAO().isUserAlreadySaved(patient.getUuid())){
```
rkorytkowski: Wouldn't it be more efficient to just create patient dao once?

openmrs-client/src/main/java/org/openmrs/mobile/activities/lastviewedpatients/LastViewedPatientsPresenter.java
```diff
@@ -62,7 +66,7 @@ public void updateLastViewedList() {
             public void onResponse(Call<Results<Patient>> call, Response<Results<Patient>> response) {
                 mLastViewedPatientsView.setSpinnerVisibility(false);
                 if (response.isSuccessful()) {
-                    mLastViewedPatientsView.updateList(response.body().getResults());
+                    mLastViewedPatientsView.updateList(filterNotDownloadedPatients(response.body().getResults()));
```
rkorytkowski: I'd rather you introduce getLastViewedNotDownloadedPatients in the restApi and unit test it.

AdamGrzybkowski: @rkorytkowski I don't really understand what you mean. Do you want to create a method in RestApi interface ? If you just want to create a method named like this in LastViewedPatientsPresenter then I'm not sure about it, because this filter has to be applied also in findPatients() method.

rkorytkowski: I see! I didn't notice the other place you used it. It's ok then.

PR #235, opened 2016-12-13
AC-242 Introduce way to unit test the presenters

AdamGrzybkowski:


openmrs-client/src/test/java/org/openmrs/mobile/test/presenters/LastViewedPatientsPresenterTest.java
```diff
@@ -0,0 +1,126 @@
...
+import retrofit2.Call;
+import retrofit2.Response;
+
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.atLeast;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class LastViewedPatientsPresenterTest {
```
rkorytkowski: Could we please create a test base class? TestBase with at least the `@RunWith` annotation and MockitoAnnotations.initMocks(this) in it.

openmrs-client/src/test/java/org/openmrs/mobile/test/MockCallOnResponse.java
```diff
@@ -0,0 +1,52 @@
+package org.openmrs.mobile.test;
+
+import java.io.IOException;
+
+import okhttp3.Request;
+import retrofit2.Call;
+import retrofit2.Callback;
+import retrofit2.Response;
+
+public class MockCallOnResponse<T> implements Call<T>{
```
rkorytkowski: How about you make it MockResultsSuccessCall<T> implements Call<Results<T>>
so that you could have a constructor like this MockResultsSuccssCall(List<T> results)...?

openmrs-client/src/test/java/org/openmrs/mobile/test/presenters/LastViewedPatientsPresenterTest.java
```diff
@@ -0,0 +1,126 @@
...
+    private ArgumentCaptor<Call<List<Patient>>> argumentCaptor;
+
+    @Mock
+    private PatientDAO patientDAO;
+    @Mock
+    private RestApi restApi;
+    @Mock
+    LastViewedPatientsContract.View view;
+
+    private LastViewedPatientsPresenter lastViewedPatientsPresenter;
```
rkorytkowski: I believe you could use @Inject here to inject mocks...

openmrs-client/src/test/java/org/openmrs/mobile/test/MockCallOnFailure.java
```diff
@@ -0,0 +1,53 @@
...
+
+
+import java.io.IOException;
+
+import okhttp3.Request;
+import retrofit2.Call;
+import retrofit2.Callback;
+import retrofit2.Response;
+
+public class MockCallOnFailure<T> implements Call<T> {
```
rkorytkowski: I'd rename it to MockFailure

openmrs-client/src/test/java/org/openmrs/mobile/test/presenters/LastViewedPatientsPresenterTest.java
AdamGrzybkowski: I did try @InjectMocks annotation but it didn't work. I believe it is due to the fact that we don't inject those objects in the presenter. We could do it the way you want but to make it possible we should start using Dagger but this is a little bit more complex task so I didn't want to do this as part of this ticket. The only way I could have mocks inside the presenter was this constructor.

PR #240, opened 2016-12-15
AC-242 applying chages accordign to post-review

AdamGrzybkowski:


openmrs-client/src/test/java/org/openmrs/mobile/test/ACBaseUnitTest.java
```diff
@@ -0,0 +1,16 @@
+package org.openmrs.mobile.test;
+
+import org.junit.Before;
+import org.junit.runner.RunWith;
+import org.mockito.MockitoAnnotations;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ACBaseUnitTest {
```
rkorytkowski: I think you should make it abstract and also it shouldn't end with Test e.g. ACUnitTestBase so that we do not have to tweak build config to not execute the class as a test.

openmrs-client/src/test/java/org/openmrs/mobile/test/MockResponse.java
```diff
@@ -0,0 +1,64 @@
...
+import okio.BufferedSource;
+import retrofit2.Call;
+import retrofit2.Callback;
+import retrofit2.Response;
+
+public class MockResponse implements Call<Results>{
+
+    Response response;
+
+    public MockResponse(List resultList, boolean successful) {
```
rkorytkowski: Let's use generics `List<T>`

rkorytkowski: Also I would split it into 2 classes MockSuccessResponse and MockErrorResponse. The thing is MockErrorResponse should let you test different error codes and possibly error messages in response body.

openmrs-client/src/test/java/org/openmrs/mobile/test/ACBaseUnitTest.java
```diff
@@ -0,0 +1,16 @@
...
+
+    @Before
+    public void mockAnnotations() {
```
rkorytkowski: Usually you call it initMocks()

openmrs-client/src/test/java/org/openmrs/mobile/test/presenters/LastViewedPatientsPresenterTest.java
```diff
@@ -49,7 +42,6 @@
 
     @Before
     public void setup(){
-        MockitoAnnotations.initMocks(this);
         lastViewedPatientsPresenter = new LastViewedPatientsPresenter(
```
rkorytkowski: Did you try `@Inject` instead?

openmrs-client/src/test/java/org/openmrs/mobile/test/ACBaseUnitTest.java
AdamGrzybkowski: yea this seems more suitable

openmrs-client/src/test/java/org/openmrs/mobile/test/presenters/LastViewedPatientsPresenterTest.java
AdamGrzybkowski: There is no annotation @Inject

openmrs-client/src/test/java/org/openmrs/mobile/test/MockResponse.java
AdamGrzybkowski: I think we should do MockSuccesResponse<T> implements Call<T> not Results because this is not the only object that is being returned by RestApi.
I agree about splitting it into 2 classes

openmrs-client/src/test/java/org/openmrs/mobile/test/presenters/LastViewedPatientsPresenterTest.java
rkorytkowski: Sorry, it's @InjectMocks

AdamGrzybkowski: So the @InjectMocks won't work unless we start using Dagger

PR #245, opened 2016-12-16
AC-256 - New validation logic in login activity

tmarzeion:


openmrs-client/src/main/java/org/openmrs/mobile/activities/dialog/CustomFragmentDialog.java
```diff
@@ -41,6 +41,8 @@
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
+import org.odk.collect.android.openmrs.provider.OpenMRSFormsProviderAPI;
+import org.odk.collect.android.openmrs.provider.OpenMRSInstanceProviderAPI;
```
AdamGrzybkowski: unused dependencies, probably better to clean that :)

openmrs-client/src/main/java/org/openmrs/mobile/activities/dialog/CustomFragmentDialog.java
```diff
@@ -52,6 +54,7 @@
 import org.openmrs.mobile.activities.registerpatient.SimilarPatientsRecyclerViewAdapter;
 import org.openmrs.mobile.application.OpenMRS;
 import org.openmrs.mobile.bundle.CustomDialogBundle;
+import org.openmrs.mobile.databases.OpenMRSSQLiteOpenHelper;
```
AdamGrzybkowski: same here :)

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginFragment.java
```diff
@@ -106,40 +106,27 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
-                editor.commit();
-            }
-        });
+        loginValidatorWatcher = new LoginValidatorWatcher(mUrl, mUsername, mPassword, mDropdownLocation, mLoginButton);
 
         mUrl.setOnFocusChangeListener(new View.OnFocusChangeListener() {
             @Override
             public void onFocusChange(View view, boolean b) {
-                if (StringUtils.notEmpty(mUrl.getText().toString())  && !view.isFocused() && urlChanged) {
+                if (StringUtils.notEmpty(mUrl.getText().toString()) && !view.isFocused() && loginValidatorWatcher.isUrlChanged() || (loginValidatorWatcher.isUrlChanged() && loginValidatorWatcher.isLocationErrorOccurred())) {
```
AdamGrzybkowski: maybe cut this line so it can fit in the view ? or extract this validation to a separate method ?

openmrs-client/src/main/java/org/openmrs/mobile/listeners/watcher/LoginValidatorWatcher.java
```diff
@@ -0,0 +1,128 @@
...
+                currentText.setTextColor(Color.RED);
+            }
+        }
+        mLoginButton.setEnabled(validateAll());
+    }
+
+    @Override
+    public void onNothingSelected(AdapterView<?> adapterView) {}
+
+    private boolean validateAll() {
```
AdamGrzybkowski: shouldn't this be named different if it returns something? e.g. isAllDataValid ?

tmarzeion:
@AdamGrzybkowski 
1. Yeah, I didn't notice that dialog, I've removed this one.
2. Fixed :)

AdamGrzybkowski:
Nice @tmarzeion I really like the way it works now :) but ... :D

There is one bug related to this.
When you change the url and try to login with wrong user credentials you get the prompt that the DB will be wiped (that is ok ) and you are redirected back to login page. When you login now with correct credentials there will be no prompt about wiping and the data will remain on the phone when it should be deleted

tmarzeion:
@AdamGrzybkowski Fixed

AdamGrzybkowski:
@tmarzeion It works, but now after wrong credentials being passed the new URL that user changed in login activity is not there and you have to type it again. I know this is due to not storing it before authenticating user but maybe we could store it only temporary ?

PR #247, opened 2016-12-17
AC-230: Login in offline mode

shivtej1505:


tmarzeion:
@shivtej1505 Did You notice merge conflicts? :)

shivtej1505:
I'll update the PR

shivtej1505:
I've updated the PR, please review.

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
```diff
@@ -81,58 +83,81 @@ public void authenticateUser(final String username, final String password, final
...
-                        loginView.sendIntentBroadcast(ApplicationConstants.CustomIntentActions.ACTION_AUTH_FAILED_BROADCAST);
+                        loginView.hideLoadingAnimation();
+                        ToastUtil.error(response.message());
                     }
+                }
+
+                @Override
+                public void onFailure(Call<Session> call, Throwable t) {
+                    loginView.hideLoadingAnimation();
+                    ToastUtil.error(t.getMessage());
```
AdamGrzybkowski: Can you delegate all ToastUtil calls to the view  ? This way it will be easier to unit test the presenters later :)

openmrs-client/src/main/java/org/openmrs/mobile/utilities/ToastUtil.java
```diff
@@ -49,6 +49,11 @@ public static void notify(String message) {
         showToast(OpenMRS.getInstance(), ToastType.NOTICE, message, Toast.LENGTH_SHORT);
     }
 
+    public static void notify(int textId) {
+        Context context = OpenMRS.getInstance();
```
AdamGrzybkowski: Why not just passing OpenMRS.getInstance() instead of creating new field? :)

shivtej1505: Actually I thought the same but I'm using `context` to fetch strings from the string resources.

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
shivtej1505: Do you mean I should call all the ToastUtils method from classes other than presenters (like LoginFragment in this case) ?

AdamGrzybkowski: yes, in this case something like loginView.showErrorToast(t.getMessage())

openmrs-client/src/main/java/org/openmrs/mobile/utilities/ToastUtil.java
AdamGrzybkowski: ohh I see now, but once you move ToastUtil to the view this method won't be necessary

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
```diff
@@ -148,33 +173,47 @@ public void saveLocationsToDatabase(List<Location> locationList, String selected
...
-                    RestServiceBuilder.changeBaseUrl(url.trim());
-                    mOpenMRS.setServerUrl(url);
-                    loginView.initLoginForm(response.body().getResults(), url);
-                    loginView.startFormListService();
-                    loginView.setLocationErrorOccurred(false);
-                } else {
-                    loginView.showInvalidURLSnackbar("Failed to fetch server's locations");
-                    loginView.setLocationErrorOccurred(true);
+
+        if (NetworkUtils.hasNetwork()) {
```
AdamGrzybkowski: Shouldn't we use NetworkUtilis.isOnline() here ? hasNetwork only checks for internet connection not the sync button

@rkorytkowski could you give some thoughts  on that ? :)

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
```diff
@@ -81,58 +83,81 @@ public void authenticateUser(final String username, final String password, final
...
+
+                @Override
+                public void onFailure(Call<Session> call, Throwable t) {
+                    loginView.hideLoadingAnimation();
+                    ToastUtil.error(t.getMessage());
+                }
+            });
+        } else {
+
+            if (mOpenMRS.getFirstLogin()) {
```
AdamGrzybkowski: I guess mOpenMRS.isSetLogin would be more suitable if you are returning boolean

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
```diff
@@ -81,58 +83,81 @@ public void authenticateUser(final String username, final String password, final
...
+                if (mOpenMRS.getUsername().equals(username) && mOpenMRS.getPassword().equals(password)) {
```
AdamGrzybkowski: isn't this line enough to decide if this specific user has been logged before ? and if yes let him login in offline mode ? 
I'm asking because I couldn't find place where you set first login to false, so unless user wasn't ever logged in to the app it will always be true

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
```diff
@@ -148,33 +173,47 @@ public void saveLocationsToDatabase(List<Location> locationList, String selected
...
-                loginView.showInvalidURLSnackbar(t.getMessage());
+                @Override
+                public void onFailure(Call<Results<Location>> call, Throwable t) {
+                    loginView.hideUrlLoadingAnimation();
+                    loginView.showInvalidURLSnackbar(t.getMessage());
+                    loginView.setLocationErrorOccurred(true);
+                }
+            });
+        } else {
+            List<Location> locations = mOpenMRS.getLocations();
```
AdamGrzybkowski: We have static method LocationDAO.getLocations() so maybe there is no need to save them again in shared prefs ? @rkorytkowski

AdamGrzybkowski:
The app crashes on startup when you are not logged in and the network connection is turned off or when you are at the login screen with network turned off and try to paste new url. Here is the stacktrace http://pastebin.com/30j6XxHL

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
rkorytkowski: Yes, we should.

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
rkorytkowski: There supposed to be a DB encryption in place, which should stop you from calling LocationDAO.getLocations() before authenticating... If it is still the case, then storing login locations elsewhere does make sense.

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
shivtej1505: No, we also need to check if the credentials entered are correct. If we don't do it, anyone can login with wrong credentials and app may lose useful data.
Second line checks if any user have ever logged in, but last line ensures that only last user with correct credentials can logged in.

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
shivtej1505: I think locations should be downloaded irrespective of the state of sync. Consider a scenerio, a new user login in for first time and sync is off, then he won't able to login. It is not intutive to first go online by tapping on cloud and then login.
Anyways, syncing should be applied on data like patient's data, not on the location which is not user dependent, it is common for every user.

shivtej1505:
@AdamGrzybkowski , I've updated the PR, have a look! :)

openmrs-client/src/main/java/org/openmrs/mobile/application/OpenMRS.java
```diff
@@ -127,6 +133,11 @@ public void setVisitTypeUUID(String visitTypeUUID) {
         editor.commit();
     }
 
+    public boolean getLogin() {
```
AdamGrzybkowski: what about something like isUserLoggedOnline() and setUserLoggedOnLine()  ?

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
AdamGrzybkowski: @rkorytkowski what do you think  ?

AdamGrzybkowski:
@shivtej1505  please resolve conflicts :)

shivtej1505:
@AdamGrzybkowski , resolved!

PR #252, opened 2016-12-19
AC-252 Rewrite code for patient highlighting

AdamGrzybkowski:


openmrs-client/src/main/java/org/openmrs/mobile/api/retrofit/PatientApi.java
```diff
@@ -96,9 +96,6 @@ public void onResponse(Call<Patient> call, Response<Patient> response) {
                                     if (response.isSuccessful()) {
                                         Patient newPatient = response.body();
 
-                                        ToastUtil.success("Patient " +patient.getPerson().getName().getNameString()
-                                                +" created with UUID " + newPatient.getUuid());
-
```
AdamGrzybkowski: I've deleted it because when syncing a lot of patients all those toast are slowing down the UI. This was happening also when downloading bigger number of patients.
I think toast are good when something goes wrong, but when all is ok we should do something different or if anything at all.

openmrs-client/src/main/java/org/openmrs/mobile/activities/lastviewedpatients/LastViewedPatientRecyclerViewAdapter.java
```diff
@@ -167,83 +125,62 @@ public PatientViewHolder(View itemView) {
...
-            }
-            else if (!value && isSelected){
-                // unselected and previously was selected, otherwise no change needed
-                    howManySelected--;
-                    holder.mRowLayout.setSelected(false);
-                    holder.mRowLayout.setBackgroundResource(R.drawable.card);
+        @Override
+        public void onClick(View view) {
+            if(isLongClicked){
+                setSelected(!mRowLayout.isSelected());
```
tmarzeion: This one is awsum

openmrs-client/src/main/java/org/openmrs/mobile/api/retrofit/PatientApi.java
tmarzeion: I agree, downloading many patients was painful, because of this endless toast

tmarzeion:
@AdamGrzybkowski It looks fine, I've tested it out and it works great :)

PR #266, opened 2016-12-27
AC-270 - Disabled display of encounters with unknown encounterType

tmarzeion:


openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitDashboardFragment.java
```diff
@@ -97,7 +100,18 @@ public void moveToPatientDashboard() {
...
     @Override
     public void updateList(List<Encounter> visitEncounters) {
-        mExpandableListAdapter = new VisitExpandableListAdapter(this.getActivity(), visitEncounters);
+
+        List<Encounter> displayableEnounters  = new ArrayList<>();
+
+        for (Encounter encounter : visitEncounters) {
+            String[] displayableEncounterTypes = ApplicationConstants.EncounterTypes.ENCOUNTER_TYPES_DISPLAYS;
+            String encounterTypeDisplay = encounter.getEncounterType().getDisplay();
+            if (Arrays.asList(displayableEncounterTypes).contains(encounterTypeDisplay)) {
```
rkorytkowski: Create collection outside the for loop to do it just once. Also it's always more efficient to use HashSet when you intend to call contains (O(1) instead of O(n)). It's a small set so it doesn't make much difference, but it's a good habit to do it right and shows you understand the difference.

tmarzeion: I've probably missed this one :) Thanks for review :)
Fixed

PR #270, opened 2017-01-05
AC-258: Edit Patient Information

ykarim:
Added Update Patient functionality using the existing Register Patient Form. 
Also changed some fields of register patient to better help users fill in information. 
Renamed RegisterPatient to PatientInfo.

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/patientinfo/PatientInfoFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/patientinfo/PatientInfoFragment.java
@@ -12,7 +12,7 @@
  * Copyright (C) OpenMRS, LLC.  All Rights Reserved.
  */
 
-package org.openmrs.mobile.activities.registerpatient;
+package org.openmrs.mobile.activities.patientinfo;
 
 import android.app.DatePickerDialog;
 import android.content.Context;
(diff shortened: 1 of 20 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/activities/patientinfo/PatientInfoFragment.java
```diff
@@ -230,6 +239,66 @@ private Patient createPatient() {
...
+                bdt = bdt.minusMonths(mondiff);
+                person.setBirthdateEstimated(true);
+                birthdate = bdt.toString();
+            }
+        } else {
+            birthdate = bdt.toString();
+        }
+        person.setBirthdate(birthdate);
+
+        Patient updatedPatient = new PatientDAO().findPatientByID(id);
```
tmarzeion: Accessing Data Sources isn't allowed in View part of MVP pattern.
Data is provided by presenter :)
I suggest You to pass whole Patient object via Presenter to Fragment, so Fragment doesn't have to access DB (Presenter does it instead). Just replace `setPatientIdBundle()` with something like `setPatientToUpdate()`.
Its harder to handle all these MVP pattern conditions, but Its worth It when it comes to code readability and code testing :)

openmrs-client/src/main/res/menu/patient_details_menu.xml
```diff
@@ -21,4 +21,8 @@
         android:title="@string/action_synchronize_patients"
         android:icon="@drawable/ic_menu_refresh"
         app:showAsAction="always"/>
+    <item android:id="@+id/actionUpdatePatient"
+        android:title="@string/action_update_patient_data"
+        android:icon="@drawable/ic_edit"
```
tmarzeion: This drawable resource is based on `*.png` assets (4 different sizes for different dpi).
Its good approach when using image assets like photos, pictures (and generally everything that is difficult to draw as `*.svg`), but when we are using simple material icon like this one, its possible to add this as vector asset, which can be resized infinitely by using only 1 file instead of 4 :)

ykarim:
@tmarzeion I've been trying to work on updating the toolbar of PatientDashboardActivity. I've added an onActivityResult() to PatientDashboardActivity, but when I check the presenter after updating the patient, I get PatientDiagnosisPresenter. Isn't the presenter supposed to be PatientDetaulsPresenter?

ykarim:
Made the changes. Wasn't able to fix the toolbar issue.

tmarzeion:
@ykarim 
I will create ticket for issue with toolbar refreshing, I will also rename `patientinfo` to something more intuitive like `addeditpatient` :)
Good job :)

ykarim:
Thanks @tmarzeion

PR #276, opened 2017-01-10
AC-280 - Fixed Add Patients button sensitivity to Sync State

tmarzeion:


openmrs-client/src/main/java/org/openmrs/mobile/activities/ACBaseActivity.java
```diff
@@ -101,31 +101,24 @@ public boolean onPrepareOptionsMenu(final Menu menu) {
...
+
+    public boolean setSyncState(boolean enabled) {
+        final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(OpenMRS.getInstance());
+        SharedPreferences.Editor editor = prefs.edit();
+        editor.putBoolean("sync", enabled);
+        editor.apply();
+        setSyncButtonState(enabled);
+        return true;
+    }
+
```
AdamGrzybkowski: @tmarzeion Could we extract these methods to OpenMRS class? This might be helpful in other places and we won't be accessing data storage directly from an activity.

tmarzeion: Good idea, I've moved these methods to `OpenMRS`

PR #280, opened 2017-01-11
AC-283: Fixed selecting patients in Download Patient

lkukielka:
https://issues.openmrs.org/browse/AC-283
AC-283: Selecting from Download Patient doesn't work after repopening app
I fixed selecting patients in Download Patients view. Now patients are selected even after reopening app.


AdamGrzybkowski:
@lkukielka I think we should wait with this until we merge AC-285

PR #282, opened 2017-01-11
AC-282 - Locked form submit button when submit is processed

tmarzeion:


tmarzeion:
@AdamGrzybkowski Yeah I've missed something while doing some pre-commit tweaks. Fixed

PR #286, opened 2017-01-12
AC-281 fixed crashing app when repeatedly pressing register or back b…

AdamGrzybkowski:
…utton


AdamGrzybkowski:
@tmarzeion PR updated

PR #288, opened 2017-01-12
AC-263: Fixed patient DOB bug

lkukielka:
AC-263: Patient DOB goes back one day when synced with server
https://issues.openmrs.org/browse/AC-263
I fixed patient DOB bug

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
@@ -36,6 +36,8 @@ import android.widget.TextView;
 
 import org.joda.time.DateTime;
 import org.joda.time.LocalDate;
+import org.joda.time.format.DateTimeFormat;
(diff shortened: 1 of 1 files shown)


AdamGrzybkowski:
@lkukielka It looks like the bug is not fixed with your changes.

openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
```diff
@@ -61,6 +63,7 @@
 
     LocalDate birthdate;
     DateTime bdt;
+    private DateTimeFormatter dtf = DateTimeFormat.forPattern("yyyy-MM-dd");
```
shivtej1505: shouldn't we give self explanatory name to variables, something like `dataTimeFmt` ?

AdamGrzybkowski: @shivtej1505  I agree. The code will be easier to read
@lkukielka please update the PR and resolve conflicts :)

PR #303, opened 2017-01-17
AC-323 DatePicker prevents future dates as DOB in registration form

yatna:
Now the user cannot select future dates as Date of Birth in the registration form, as they have been disabled. Here is a screen-shot

screenshot_20170117-200015 1


AdamGrzybkowski:
@yatna I've missed that but please always start your commit message with issue id e.g AC-323

PR #305, opened 2017-01-18
AC-329 List of countries don't show automatically

DefCon-007:
Now when the user taps "Country" field in Register Patient the country suggestions are shown even if the user doesn't type anything.

screenshot_20170118-133906

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
@@ -54,6 +54,7 @@ import org.openmrs.mobile.utilities.ToastUtil;
 import org.openmrs.mobile.utilities.ViewUtils;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.List;
 
@@ -88,6 +89,8 @@ public class AddEditPatientFragment extends Fragment implements AddEditPatientCo
     private TextView countryerror;
 
     private Button submitConfirm;
+
+    private String[] countries;
 
 
     @Override
@@ -354,7 +357,7 @@ public class AddEditPatientFragment extends Fragment implements AddEditPatientCo
     }
 
     private void addSuggestionsToAutoCompleTextView() {
-        String[] countries = getContext().getResources().getStringArray(R.array.countries_array);
+        countries = getContext().getResources().getStringArray(R.array.countries_array);
         ArrayAdapter<String> adapter = new ArrayAdapter<>(getContext(),
                 android.R.layout.simple_dropdown_item_1line, countries);
         edcountry.setAdapter(adapter);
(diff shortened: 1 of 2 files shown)


tmarzeion:
Hey @DefCon-007 :)
Its great that You want to contribute Your dev knowledge into OpenMRS :)
Actually we don't need feature like this one, since we want to keep `country` field as textfield with autocomplete instead of more spinner-like look.
When adding new feature, You have to create ticket and **wait until it is accepted** and moved from backlog, usually it doesn't take much time to mark issue as `ready for work` (approx 1 day).

PS. I think this feature would be nice if it only worked when length of value typed in `country` field is `>= 3`, so we would avoid spinner-like behavior when clicking on empty field, but we would have nice autocomplete when changing focus into `country` field.
Also notice that we wouldn't like to show autocomplete when for example `Algeria` is already typed into `country` field.
PPS. There is a bug which makes autocomplete suggest incorrect countries. To reproduce:
- Type `Al` into `country` field
- Choose `Algeria`
- Click on field again (both `Albania` and `Algeria` is shown)

DefCon-007:
Hey @tmarzeion  , thanks for pointing it out, I will wait for the issue to get accepted next time.
I will make the autocomplete to show suggestions only when the length of value typed in the country field is `>= 3` . I will also try to fix the bug you pointed out.

P.S. : I will also update the issue on JIRA.

DefCon-007:
@tmarzeion : Made the necessary changes

tmarzeion:
Hey @DefCon-007 :)
We want behavior such as:
1. Type `Pol` in `country` field
2. Don't choose country, click on another `EditText` instead
3. Click on `country` field - now there should be autocomplete visible with `Poland` to choose :)

At this moment list of countries isn't displayed automatically, which is main subject of this issue :)

PS. I've tested that out and I think that we can go back to autocomplete threshold to `2` instead of `3` :)

DefCon-007:
Hey @tmarzeion made the requested changes.

tmarzeion:
@DefCon-007:
We don't want to display autocomplete when name of country is correct :)

To reproduce:
- Type `Poland` into `country` field
- Click on another `EditText`
- Click back on `country` field
- `Poland` in autocomplete is still visible.

DefCon-007:
@tmarzeion : 
Autocomplete is not shown now if the country name is correct.Though I have not made it case-insensitive.
For example 

- If `Poland` is in the text field suggestions will not be shown.
- But if `poland` is written `Poland` will be suggested.

Should I make it case-insensitive?

AvijitGhosh82:
@DefCon-007  You can automatically make the first letter capital and to a case insensitive search on the list of countries.

tmarzeion:
@DefCon-007 this works just fine, but You can add feature mentioned by @AvijitGhosh82 - `automatically make the first letter of the edittext capital` would be pretty neat :)

DefCon-007:
@AvijitGhosh82 Made all the requested changes

AvijitGhosh82:
@DefCon-007 You pushed the change I requested to AC-330, you need to add it here :)

PR #307, opened 2017-01-20
AC-326 Text color is preserved during orientation change in Form Entry

yatna:
In all the form categories under 'Form Entry', the highlighted text color (red when entry is out of range) is preserved during orientation change. Here are a few screen shots- 
screenshot_20170121-032821
screenshot_20170121-033530

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/utilities/InputField.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/utilities/InputField.java
@@ -24,6 +24,7 @@ public class InputField implements Serializable, Parcelable {
     private int id;
     private String concept;
     private double value = -1.0;
+    private boolean isRed = false;
 
     public InputField(String concept) {
         this.concept = concept;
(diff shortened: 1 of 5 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/utilities/InputField.java
```diff
@@ -14,6 +14,7 @@
 
 package org.openmrs.mobile.utilities;
 
+import android.graphics.Color;
```
AvijitGhosh82: @yatna You'd need to take care of lint issues.

openmrs-client/src/main/java/org/openmrs/mobile/utilities/InputField.java
```diff
@@ -24,6 +25,7 @@
     private int id;
     private String concept;
     private double value = -1.0;
+    private int isRed = 0;
```
AdamGrzybkowski: @yatna Why int when you only use 0 or 1 ? Wouldn't boolean be better here  ?

openmrs-client/src/main/java/org/openmrs/mobile/utilities/InputField.java
yatna: Sure, working on it.

PR #310, opened 2017-01-25
AC-331 Problem in editing patient's Vital Form

AdamGrzybkowski:
I have added TextView above each numeric field. See below
screenshot_1485433580


AdamGrzybkowski:
@shivtej1505 I agree this looks strange to have it twice but this is rather how this form is created(other forms seems fine?) or bug in creating layout that is not part of this ticket. But I think it is the option nr 1

AvijitGhosh82:
@shivtej1505 I was wondering the same thing when I was first designing this page, and it would appear that the form happens to have the form title and page title same. If the form had other pages this would not be an issue. 

@AdamGrzybkowski  I'd suggest to add a bit of padding to the left of the question texts so that it looks neater.

AdamGrzybkowski:
@AvijitGhosh82 I've increased left margin and replaced screenshot above with the new one

tmarzeion:
@AdamGrzybkowski Doubled `vitals` label is caused by form resource, not android app itself :hamburger:

PR #312, opened 2017-01-26
AC-326 Text color is preserved during orientation change in Form Entry

yatna:
This is a fresh PR in reference to #307 . I closed the previous one as I messed up with the git while editing it. I have made the changes as suggested in #307.

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayPageFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayPageFragment.java
@@ -81,6 +81,12 @@ public class FormDisplayPageFragment extends Fragment implements FormDisplayCont
         if (savedInstanceState != null) {
             FormFieldsWrapper formFieldsWrapper = (FormFieldsWrapper) savedInstanceState.getSerializable(ApplicationConstants.BundleKeys.FORM_FIELDS_BUNDLE);
             inputFields = formFieldsWrapper.getInputFields();
+            for(InputField field:inputFields){
+                if(field.isRed()){
+                    RangeEditText ed = (RangeEditText) getActivity().findViewById(field.getId());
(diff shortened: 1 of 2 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/utilities/InputField.java
```diff
@@ -64,12 +70,14 @@ public void writeToParcel(Parcel dest, int flags) {
...
         dest.writeString(this.concept);
         dest.writeDouble(this.value);
+        dest.writeInt(this.isRed?1:0);
     }
 
     protected InputField(Parcel in) {
         this.id = in.readInt();
         this.concept = in.readString();
         this.value = in.readDouble();
+        this.isRed=(in.readInt()==1)?true:false;
```
tmarzeion: `?true:false` is redundant, since when `in.readInt()` is `true` it is `true` (It may sound silly but think about it :rofl:)

openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayPageFragment.java
```diff
@@ -315,10 +321,14 @@ public LinearLayout createSectionLayout(String sectionLabel) {
     public List<InputField> getInputFields() {
         for (InputField field:inputFields) {
             RangeEditText ed=(RangeEditText) getActivity().findViewById(field.getId());
-            if(!isEmpty(ed))
+            if(!isEmpty(ed)){
                 field.setValue(Double.parseDouble(ed.getText().toString()));
-            else
+                boolean isRed=(ed.getCurrentTextColor()==ContextCompat.getColor(OpenMRS.getInstance(), R.color.red))?true:false ;
```
tmarzeion: `?true:false` is redundant

yatna:
@tmarzeion Changes made. Thanks for the suggestions :-)

tmarzeion:
@yatna That looks much better now! :)
Thanks for contribution! :)
Merging

PR #313, opened 2017-01-28
AC-336 : Turn auto-lock off for forms

DefCon-007:
Along with adding the required feature I have also removed one unnecessary print statement.


AvijitGhosh82:
Ah all right. Merging :)

PR #314, opened 2017-01-28
AC-304: Lint adjustment - Usability

ykarim:
Not exactly sure what to do with the AppIndexing lint error.


tmarzeion:
@ykarim Its ok, we don't need that :)
Thanks for help :)

PR #316, opened 2017-01-30
AC-334 : Numeric fields validation not working when orientation changed

rhari991:
Moved the storage of form fragments to FormPageAdapter.

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayMainPresenter.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayMainPresenter.java
@@ -12,6 +12,7 @@ package org.openmrs.mobile.activities.formdisplay;
 
 import android.app.Fragment;
 import android.os.Bundle;
+import android.util.SparseArray;
 
 import org.joda.time.LocalDateTime;
 import org.openmrs.mobile.api.EncounterService;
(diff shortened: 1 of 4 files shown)


AdamGrzybkowski:
@rhari991 Please add ticket id to the commit message and claim this ticket on JIRA

openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayMainPresenter.java
```diff
@@ -67,10 +68,10 @@ public void createEncounter() {
 
         List<Obscreate> observations=new ArrayList<>();
 
-        List<Fragment> activefrag=getActiveFragments();
+        SparseArray<Fragment> activefrag = mPageAdapter.getRegisteredFragments();
         boolean valid=true;
-        for (Fragment f:activefrag) {
-            FormDisplayPageFragment formPageFragment=(FormDisplayPageFragment)f;
+        for (int i = 0;i < activefrag.size();i++) {
```
tmarzeion: Why did You replace `(Fragment f:activefrag)` with `(int i = 0;i < activefrag.size();i++)`?

rhari991:
@AdamGrzybkowski done :)

openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayMainPresenter.java
rhari991: I've used a SparseArray to store the Fragments, and the for each loop does not let you iterate through it.

tmarzeion: Ok, I guess that its because `SparseArray` is extending `Object` directly and it isn't typical Java collection at all :)
Its OK then :)

tmarzeion:
@rhari991 Great work, it looks like You know Your job :smile: 
Thanks for contribution :)
Merging
PS. Moonkin OP :owl:

PR #319, opened 2017-02-02
AC-343 Infinite Scroll functionality for Last Viewed Patients Activity

AdamGrzybkowski:
ezgif com-video-to-gif

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/lastviewedpatients/LastViewedPatientRecyclerViewAdapter.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/lastviewedpatients/LastViewedPatientRecyclerViewAdapter.java
@@ -25,6 +25,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.CheckBox;
 import android.widget.LinearLayout;
+import android.widget.ProgressBar;
 import android.widget.TextView;
 
 import org.openmrs.mobile.R;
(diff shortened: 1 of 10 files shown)


AdamGrzybkowski:
@shivtej1505 Thanks for checking it out. I took another look and found two bugs in this.
1. When most of the first fetched patients are saved locally they won't be shown and you would already see progress bar at the bottom. I fixed it by doing a check how many patients will be shown, if not enough then I'm making another request and add those to the previous one.
2. On screen orientation new Presenter is created  which means startIndex field will be 0 but the list will have some patients. So after scrolling down it would fetch same patients. This is also fixed by saving startIndex in bundle.

openmrs-client/src/main/java/org/openmrs/mobile/activities/lastviewedpatients/LastViewedPatientsPresenter.java
```diff
@@ -72,15 +79,29 @@ public void start() {
...
         call.enqueue(new Callback<Results<Patient>>() {
             @Override
             public void onResponse(Call<Results<Patient>> call, Response<Results<Patient>> response) {
                 if (response.isSuccessful()) {
-                    mLastViewedPatientsView.updateList(filterNotDownloadedPatients(response.body().getResults()));
-                    setViewAfterPatientDownloadSuccess();
-                    mLastViewedPatientsView.enableSwipeRefresh(true);
+                    setStartIndexIfMorePatientsAvailable(response.body().getLinks());
+                    patients.addAll(filterNotDownloadedPatients(response.body().getResults()));
+                    if (patients.size() < 7 && !isDownloadedAll) {
```
tmarzeion: This magic number should be constant field with purpose described in its name.
See: [Magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) - `Unique values with unexplained meaning or multiple occurrences which could (preferably) be replaced with named constants` (Its one of meaning)

openmrs-client/src/main/java/org/openmrs/mobile/activities/lastviewedpatients/LastViewedPatientRecyclerViewAdapter.java
```diff
@@ -112,14 +142,16 @@ private boolean isPatientSelected(int position) {
     }
 
     @Override
-    public void onViewDetachedFromWindow(PatientViewHolder holder) {
-        holder.clearAnimation();
+    public void onViewDetachedFromWindow(RecyclerView.ViewHolder holder) {
+        if (holder instanceof PatientViewHolder) {
```
AdamGrzybkowski: Yes, there is ProgressBarViewHolder... see below

openmrs-client/src/main/java/org/openmrs/mobile/activities/lastviewedpatients/LastViewedPatientsPresenter.java
AdamGrzybkowski: @tmarzeion done

PR #326, opened 2017-02-03
AC-342 :Implement RxJava in EncounterDAO

rhari991:


AdamGrzybkowski:
@rhari991 Remember to change issue status on JIRA to Request code review

rhari991:
My bad, forgot to do that.

AdamGrzybkowski:
@rhari991 What about other methods form EncounterDAO ?

rhari991:
@AdamGrzybkowski I wasn't sure if I should modify them or not, since they are not used in presenters and require immediate returns. Should I use toBlocking().first() for them ?

AdamGrzybkowski:
That's a good question to be honest. If they are not blocking the main thread then we probably don't need that but on the other hand the Dao interface will be inconsistent which might be confusing... @rkorytkowski what do you think ?

rkorytkowski:
I wouldn't push towards making interface methods consistently return Observable if there's no plan to make use of it and simply call toBlocking().first().

Moving forward let's switch to rxjava just in the places, which seem to be unresponsive for the user and make sense to be executed asynchronously.

AdamGrzybkowski:
@rhari991 Can you change this to mach subscribe/unsubsribe pattern you introduced ?

rhari991:
@AdamGrzybkowski done

PR #328, opened 2017-02-05
AC 374 - Instant warning on entering values beyond specified ranges in forms

yatna:
Earlier, if the user enters values if forms which are out of range, toast messages are given only after pressing the submit button. Now, instant warning messages to be given as soon as a value is entered beyond the specified range.

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayPageFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayPageFragment.java
@@ -17,6 +17,7 @@ import android.support.v4.content.ContextCompat;
 import android.text.InputType;
 import android.util.TypedValue;
 import android.view.Gravity;
+import android.view.KeyEvent;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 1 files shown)


AdamGrzybkowski:
@yatna Thanks for your PR. 
Couple of things to change:
1. This check is not executed when deleting back the values. The error icon disappears when the value is still wrong.
2. After clicking submit text of fields with wrong value is red and is not changed to black if set to correct value. This could be done in your listener too. 
3. I would remove Toast that are shown after clicking submit button when some values are out of range. I think the warning you provided are enough. Of course we cannot allow to submit with wrong values so instead of showing Toast, hide keyboard if shown and scroll to top.

openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayPageFragment.java
```diff
@@ -114,6 +115,22 @@ public void createAndAttachNumericQuestionEditText(Question question, LinearLayo
...
             ed.setLowerlimit(Double.parseDouble(question.getQuestionOptions().getMin()));
+            ed.setOnKeyListener(new View.OnKeyListener() {
+                @Override
+                public boolean onKey(View v, int keyCode, KeyEvent event) {
+                    double max = Double.parseDouble(question.getQuestionOptions().getMax());
+                    double min = Double.parseDouble(question.getQuestionOptions().getMin());
+                    double value= -1;
+
+                    if(!ed.getText().toString().trim().equals("")){
+                        value =Double.parseDouble(ed.getText().toString().trim());
```
tmarzeion: `ed.getText().toString().trim()` can be extracted to variable so it could be reused in condition above. It also will give some more clue to developer that will look into this code  about what this condition does :smiley:

openmrs-client/src/main/java/org/openmrs/mobile/activities/formdisplay/FormDisplayPageFragment.java
```diff
@@ -114,6 +115,22 @@ public void createAndAttachNumericQuestionEditText(Question question, LinearLayo
...
+                    }
+                    if(value != -1 && (value < min || value > max)){
```
tmarzeion: I think its possible to just replace `value != -1 && (value < min || value > max)` with `!(value >= min && value <= max)`
which would just mean `if value doesn't fit`

AdamGrzybkowski:
ping @yatna

AdamGrzybkowski:
@yatna are you still working on this ?

tmarzeion:
Closed since yatna is inactive for 2 weeks

PR #335, opened 2017-02-09
AC-358 Unit test - LoginPresenter

AdamGrzybkowski:


openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginPresenter.java
```diff
@@ -88,16 +109,15 @@ public void authenticateUser(final String username, final String password, final
...
         if (NetworkUtils.isOnline()) {
             mWipeRequired = wipeDatabase;
             RestApi restApi = RestServiceBuilder.createService(RestApi.class, username, password);
             Call<Session> call = restApi.getSession();
             call.enqueue(new Callback<Session>() {
                 @Override
                 public void onResponse(Call<Session> call, Response<Session> response) {
-                    mLogger.d(response.body().toString());
                     if (response.isSuccessful()) {
+                        mLogger.d(response.body().toString());
```
shivtej1505: I think if we move this line inside `if` condition then we won't get any information on the log file in case if there is any error.

AdamGrzybkowski: I agree but I did that because i was getting NPE. I will try to figure it out if this was only problem with mocked responsebody

AdamGrzybkowski: I'm almost sure I cannot change this. The body is set to null when creating error response by retrofit class. I did confirm this by testing in practice, so this could crash the app

AdamGrzybkowski:
@shivtej1505 @tmarzeion feel free to merge this if you think it's done well

PR #336, opened 2017-02-09
AC-359 - Unit test - SyncedPatientsPresenter

tmarzeion:


openmrs-client/src/test/java/org/openmrs/mobile/test/presenters/SyncedPatientsPresenterTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private List<Patient> patientList;
+
+    @Before
+    public void setUp() {
+        syncedPatientsPresenter = new SyncedPatientsPresenter(view, patientDAO);
+
+        patientList = new ArrayList<>();
+        Patient patient1 = createPatient(1L);
+        patient1.getPerson().getName().setGivenName("TEST");
```
AdamGrzybkowski: What about changing createPatient() to set name based on id passed to this method eg "patient_{id}"? This way we can reuse this later :)

openmrs-client/src/test/java/org/openmrs/mobile/test/presenters/SyncedPatientsPresenterTest.java
```diff
@@ -0,0 +1,80 @@
...
+        syncedPatientsPresenter.updateLocalPatientsList();
+        verify(view).updateListVisibility(true);
+    }
+
+    @Test
+    public void shouldShowListWhenFoundPatientsByQuery() {
+        when(patientDAO.getAllPatients()).thenReturn(patientList);
+        syncedPatientsPresenter.setQuery("TEST");
+        syncedPatientsPresenter.updateLocalPatientsList();
+        verify(view).updateListVisibility(true);
```
AdamGrzybkowski: why you are not verify updateAdapter(patientList) method ? You can check if it was called with the proper list

openmrs-client/src/test/java/org/openmrs/mobile/test/presenters/SyncedPatientsPresenterTest.java
```diff
@@ -0,0 +1,88 @@
...
+        verify(view).updateAdapter(patientList);
+    }
+
+    @Test
+    public void shouldShowListWhenFoundPatientsByQuery() {
+        final String query = "Patient_3";
+        List<Patient> filteredPatientList = FilterUtil.getPatientsFilteredByQuery(patientList, query);
+        when(patientDAO.getAllPatients()).thenReturn(Observable.just(patientList));
```
AdamGrzybkowski: @tmarzeion  What about not using FilterUtil here, so you can also test if the list is filtered correctly ? In this case we know we should have the singleton list containing patient3 and this is what we should verify.

PR #340, opened 2017-02-11
AC-375: We need to get rid of compressing patient photo because of qu…

shivtej1505:
…ality loss

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/databases/DBOpenHelper.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/databases/DBOpenHelper.java
@@ -382,7 +382,7 @@ public class DBOpenHelper extends OpenMRSSQLiteOpenHelper {
     }
     private byte[] bitmapToByteArray(Bitmap image) {
         ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-        image.compress(Bitmap.CompressFormat.JPEG, 100, outputStream);
+        image.compress(Bitmap.CompressFormat.PNG, 0, outputStream);
         return outputStream.toByteArray();
     }
 }
--- a/openmrs-client/src/main/java/org/openmrs/mobile/models/Person.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/models/Person.java
@@ -15,6 +15,8 @@ import android.graphics.Bitmap;
 import com.google.gson.annotations.Expose;
 import com.google.gson.annotations.SerializedName;
 
+import org.openmrs.mobile.utilities.ImageUtils;
+
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
@@ -170,6 +172,10 @@ public class Person extends Resource implements Serializable {
 
     public Bitmap getPhoto() {
         return photo;
+    }
+
+    public Bitmap getResizedPhoto() {
+        return ImageUtils.resizePhoto(this.photo);
     }
 
     public void setPhoto(Bitmap patientPhoto) {
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
@@ -22,7 +22,6 @@ import android.content.Intent;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Color;
-import android.graphics.Matrix;
 import android.media.ExifInterface;
 import android.media.ThumbnailUtils;
 import android.net.Uri;
@@ -47,7 +46,6 @@ import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.ProgressBar;
 import android.widget.RadioGroup;
-import android.widget.RelativeLayout;
 import android.widget.ScrollView;
 import android.widget.TextView;
 
@@ -69,6 +67,7 @@ import org.openmrs.mobile.models.PersonAddress;
 import org.openmrs.mobile.models.PersonName;
 import org.openmrs.mobile.utilities.ApplicationConstants;
 import org.openmrs.mobile.utilities.DateUtils;
+import org.openmrs.mobile.utilities.ImageUtils;
 import org.openmrs.mobile.utilities.StringUtils;
 import org.openmrs.mobile.utilities.ToastUtil;
 import org.openmrs.mobile.utilities.ViewUtils;
@@ -126,8 +125,9 @@ public class AddEditPatientFragment extends ACBaseFragment<AddEditPatientContrac
     private String[] countries;
     private ImageView patientImageView;
 
-    private FloatingActionButton capturePhoto;
+    private FloatingActionButton capturePhotoBtn;
     private Bitmap patientPhoto = null;
+    private Bitmap resizedPatientPhoto = null;
     private String patientName;
     private File output = null;
     private final static int IMAGE_REQUEST = 1;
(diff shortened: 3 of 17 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/databases/DBOpenHelper.java
```diff
@@ -347,7 +347,7 @@ public Long insertLocation(SQLiteDatabase db, Location loc) {
     }
     private byte[] bitmapToByteArray(Bitmap image) {
         ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-        image.compress(Bitmap.CompressFormat.JPEG, 100, outputStream);
+        image.compress(Bitmap.CompressFormat.PNG, 0, outputStream);
```
AdamGrzybkowski: what about the same compression in PatientPhoto class ? I think this one is crucial to. There are two more compression but I thinks those are that destroys the quality. 

Could you also set photo size to some fixed  value ? We don't need to send such big photos to server

AdamGrzybkowski: ping @shivtej1505

shivtej1505: @AdamGrzybkowski , yeah, working on it.

shivtej1505: @AdamGrzybkowski , PR updated. I removed the JPEG compression from the `PatientPhoto` class also.

openmrs-client/src/main/java/org/openmrs/mobile/models/Person.java
```diff
@@ -173,6 +173,19 @@ public Bitmap getPhoto() {
     }
 
     public void setPhoto(Bitmap patientPhoto) {
-        this.photo = patientPhoto;
+        this.photo = resizePhoto(patientPhoto);
+    }
+
+    private Bitmap resizePhoto(Bitmap photo) {
```
shivtej1505: @AdamGrzybkowski , yes. I'm doing same, I'm setting the photo in `createPerson` method in `AddEditPatientFragment`. Then newly created patient sent to `patientApi` for uploading. So it's gets resize before uploading.

AdamGrzybkowski: @shivtej1505  yes, but the resize method is called always when setting photo to Person object and this is done in some other places which means the the photo is being resized more then once ?

shivtej1505: @AdamGrzybkowski  yeah, it can be called many times but once its size is `<= 500x500 px` then it won't resize again

openmrs-client/src/main/java/org/openmrs/mobile/models/Person.java
```diff
@@ -173,6 +173,19 @@ public Bitmap getPhoto() {
...
+        double HEIGHT = 500;
+        double WIDTH = 500;
+        int height = photo.getHeight();
+        int width = photo.getWidth();
+        double aspectRatio = Math.min(HEIGHT/height, WIDTH/width);
+
+        if (0 < aspectRatio && aspectRatio < 1) {
```
shivtej1505: @AdamGrzybkowski , see here. If size is greater than required size then the photo will be resized, otherwise it won't. :)

AdamGrzybkowski: hmm then there is a different problem. If we fetch patient with photo not taken by the android app and then update patient, we will change the photo size (if it was bigger) what I think should not happen. Can't we just resize the photo after taking it ? It would be more intuitive ? 
Maybe we should make something like PhotoUtil class and do all resizing/rotating stuff there ?

shivtej1505: @AdamGrzybkowski , yeah you are right, but the problem with this approach is that if the patient photo is too large then it will take up too much memory on the device and the app may crash. This problem poped up when I was working on this PR. Maybe we can resize such large photos for device only but we won't upload them to server. What do you think ?

AdamGrzybkowski: @shivtej1505 so we have to cases:
1. resize aftter taking and before sending to server
2. resize before displaying the photo on the screen 

But we don't want to store resized picture (downloaded from remote server) in the DB.
Was this too much memory error happening when you were opening new Activity with photo passed in the bundle ?

shivtej1505: @AdamGrzybkowski , yup. Heap grows upto 30-35 MB and that's why app crashed. We have to find a way around it.

AdamGrzybkowski: @shivtej1505 Have you changed anything  ?

shivtej1505: @AdamGrzybkowski , no. I didn't change anything.

shivtej1505: @AdamGrzybkowski , so we will be saving resized photo if it is taken from the app otherwise we will save original photo, right ?

AdamGrzybkowski: Yes this is how I think we should do. Could we somehow check before saving size of the image to not crash the app ?

shivtej1505: @AdamGrzybkowski  maybe we can write a function named 'resizePhoto' which will give us resized photo for display, but it won't change original photo. We can also use to resize photos taken from the app. Sounds good ?

AdamGrzybkowski: @shivtej1505  Sounds great, but I would still resize photo taken with the mobile device before posting. It will take less space on the remote server.

shivtej1505: @AdamGrzybkowski I forgot to include that but it was in my mind. So I'll go ahead and implement it :)

shivtej1505:
@AdamGrzybkowski , I implemented the above functionality but getting some error and I couldn't resolve them. Here is the log file: http://pastebin.com/JVjU9mcv 
Please look into this, thanks ;)

AdamGrzybkowski:
@shivtej1505 It looks like you are getting an error when using rxJava somewhere. When is this happening ? e.g  when retrieving patient form DB ?

shivtej1505:
@AdamGrzybkowski , yup. I'm getting error when I try to retrieve patient from database.

AdamGrzybkowski:
@shivtej1505 I'm testing this and it seems to work fine. I do not have any errors. I've tested it for downloaded and registered via app patients.

shivtej1505:
@AdamGrzybkowski , could you please download patient name "Aagam Shah" and let me know the results. I have uploaded a large size photo for him on devtest04 server.

AdamGrzybkowski:
@shivtej1505 Works fine, no errors. I'm using Samsung S7, API 23. I'm not able to run the emulator right now so this is the only device I can test it on.

AdamGrzybkowski:
@shivtej1505 I've tested this on the emulator and it works fine

shivtej1505:
@AdamGrzybkowski , I've also tested the PR on couple of devices and getting same error. I tested on Samsung galaxy core II API 19 and Samsung galaxy J2 API 23. We must resolve this issue before merging this PR to master.
Does the photo get resized on server when you sync any patient without updating it ?

AvijitGhosh82:
Hey @shivtej1505 did you check the library I linked you to?

shivtej1505:
@AvijitGhosh82 , sorry. I was busy with my mid exams. BTW, Adam also mentioned that he wasn't getting any error, could you please check if you're getting any error or not ?

shivtej1505:
@AvijitGhosh82 , could you please check this PR and let me know if you're facing any error ? The library you pointed to is useful to resize image but the actual problem here is how to store large size image in database which is directly coming from server.

dkayiwa:
@shivtej1505 @AvijitGhosh82 do the merge conflicts on this mean it was already merged through somewhere else? 😊

shivtej1505:
@dkayiwa , no. This conflict is just due to changes at same place. @AvijitGhosh82 , please review. I have also removed unused imports in the project.

dkayiwa:
@AvijitGhosh82 do you think this is ready for merging?

AvijitGhosh82:
Hey @shivtej1505 I thought you were still working on this issue, apologies. Will check in a bit.

PR #345, opened 2017-02-16
AC-320 - Fixed options menu item text size in settings

tmarzeion:


openmrs-client/src/main/java/org/openmrs/mobile/activities/settings/SettingsActivity.java
```diff
@@ -27,12 +27,13 @@
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_settings);
-        Toolbar toolbar = (Toolbar) findViewById(R.id.toolbar);
-        if (toolbar != null) {
-            setSupportActionBar(toolbar);
+
+        ActionBar actionBar = getSupportActionBar();
+        if (actionBar != null) {
+            getSupportActionBar().setElevation(0);
```
AdamGrzybkowski: @tmarzeion  why calling a  method instead of using actionBar ?

tmarzeion: @AdamGrzybkowski Sorry, copypaste syndrome

PR #351, opened 2017-02-20
AC-374 - Added dynamic form fields validation

tmarzeion:
I didn't use `RxBinding` `filter` feature since I needed `else` logic here.


AdamGrzybkowski:
Delete Toasts and merge :)

PR #353, opened 2017-02-22
AC-381 - Replaced Dialog when authentication fails with Snackbar

tmarzeion:


openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginFragment.java
```diff
@@ -333,12 +328,20 @@ public void startFormListService() {
...
+        snackbar.show();
+    }
+
+    @Override
+    public void showInvalidLoginOrPasswordSnackbar() {
+        Snackbar snackbar = Snackbar
+                .make(mRootView, getResources().getString(R.string.invalid_login_or_password_message), Snackbar.LENGTH_LONG)
+                .setAction("EDIT", view -> {
+                    mPassword.requestFocus();
+                    mPassword.selectAll();
```
AdamGrzybkowski: Have you heard about DRY ? :nerd_face:  Maybe introduce a method that will create snackbar and return it so we don't write same code twice and more.

tmarzeion: :hamburger:

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginFragment.java
```diff
@@ -331,16 +326,32 @@ public void startFormListService() {
...
-                .make(mRootView, message, Snackbar.LENGTH_LONG)
-                .setAction("EDIT", new View.OnClickListener() {
-                    @Override
-                    public void onClick(View view) {
-                        mUrl.requestFocus();
-                        mUrl.selectAll();
-                    }
-                });
-        snackbar.show();
+        String action = "EDIT";
```
AdamGrzybkowski: @tmarzeion Shouldn't this be in string resources  ?

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginFragment.java
```diff
@@ -331,16 +326,32 @@ public void startFormListService() {
...
+        final int length = Snackbar.LENGTH_LONG;
```
AdamGrzybkowski: You are using this twice so maybe rename method (e.g. createSnackbarLong()) and set this inside it?

PR #354, opened 2017-02-22
AC-382 Sync can be turned on even if there is no network

AdamGrzybkowski:


robbyoconnor:
Shouldn't this be done via a Receiver? 

Rationale: You can set the state back and re-enable when the network returns.

AdamGrzybkowski:
@robbyoconnor  I'm not sure if setting online mode automatically when the network is back without user knowing about this would be a proper way.

robbyoconnor:
Maybe make an option to do so? @AdamGrzybkowski??

AdamGrzybkowski:
@robbyoconnor This would still require some more work then just listening for network connection. We would have to  check whether we are connected to the correct wifi network. Maybe create an issue for that  and let's discuss it there with other  people that might be more experienced in securing app then me?

robbyoconnor:
Just bringing it up :)

AvijitGhosh82:
When I'd first implemented this, it used a network receiver to auto set online mode. Raff pointed out that even if we are connected to the internet we may not be connected to the hosted/local server. 

A check for that is yet to be implemented. @AdamGrzybkowski we should ideally open an issue for this.

PR #363, opened 2017-03-13
AC-392:Download patient option not switching instantly

DefCon-007:
JIRA issue link : https://issues.openmrs.org/browse/AC-392

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/syncedpatients/SyncedPatientsActivity.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/syncedpatients/SyncedPatientsActivity.java
@@ -14,6 +14,7 @@
 
 package org.openmrs.mobile.activities.syncedpatients;
 
+import android.content.Intent;
 import android.os.Bundle;
 import android.support.v4.view.MenuItemCompat;
 import android.support.v7.app.ActionBar;
@@ -23,6 +24,7 @@ import android.view.MenuItem;
 
 import org.openmrs.mobile.R;
 import org.openmrs.mobile.activities.ACBaseActivity;
+import org.openmrs.mobile.activities.lastviewedpatients.LastViewedPatientsActivity;
 import org.openmrs.mobile.application.OpenMRS;
 import org.openmrs.mobile.utilities.ApplicationConstants;
 import org.openmrs.mobile.utilities.StringUtils;
@@ -32,6 +34,9 @@ public class SyncedPatientsActivity extends ACBaseActivity {
     private SyncedPatientsPresenter mPresenter;
     private SearchView searchView;
     private String query;
+
+    //Menu Items
+    private MenuItem mAddPatientMenuItem;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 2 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/activities/ACBaseActivity.java
```diff
@@ -49,7 +49,7 @@
     protected final OpenMRSLogger mOpenMRSLogger = mOpenMRS.getOpenMRSLogger();
     protected AuthorizationManager mAuthorizationManager;
     protected CustomFragmentDialog mCustomFragmentDialog;
-    private MenuItem mSyncbutton;
+    private MenuItem mSyncbutton,mAddPatientMenuItem;
```
tmarzeion: I think we should use new line for each field (I think it increases readability)

openmrs-client/src/main/java/org/openmrs/mobile/activities/ACBaseActivity.java
```diff
@@ -83,12 +83,16 @@ public boolean onCreateOptionsMenu(Menu menu) {
     @Override
     public boolean onPrepareOptionsMenu(final Menu menu) {
         mSyncbutton = menu.findItem(R.id.syncbutton);
+        mAddPatientMenuItem = menu.findItem(R.id.actionAddPatients);
```
tmarzeion: Its really bad behavior to add `mAddPatientMenuItem` in this class.
`ACBaseActivity` is not intended to handle behavior like this and it should be solved by using some listener between `mAddPatientMenuItem` to `mSyncbutton` inside `SyncedPatientsFragment` - not `ACBaseActivity`.

openmrs-client/src/main/java/org/openmrs/mobile/activities/ACBaseActivity.java
```diff
@@ -138,6 +145,12 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
     }
 
+    private void enableAddPatient(boolean enabled) {
+        int resId = enabled ? R.drawable.ic_add : R.drawable.ic_add_disabled;
+        mAddPatientMenuItem.setEnabled(enabled);
```
tmarzeion: Gives `NPE` when using sync button in main dashboard (or anywhere else where `mAddPatientMenuItem` isn't defined)
Dont check if `mAddPatientMenuItem` is null to workaround this. Get rid of `mAddPatientMenuItem` from `ACBaseActivity` instead.

DefCon-007:
@tmarzeion I tried to use the `onOptionsItemSelected` in `SyncedPatientsFragment` but for some reasons when I click the sync button it is not registered by the `SyncedPatientsFragment`

tmarzeion:
@DefCon-007 
I've moved menu logic to Activity. It may be a bit hacky when using MVP, but I think its easiest solution :)
Do it similar to https://github.com/openmrs/openmrs-contrib-android-client/pull/365 :)

DefCon-007:
@AvijitGhosh82 @tmarzeion For some reasons now in the `SyncedPatientsActivity` back button is not working.

PR #371, opened 2017-03-22
AC-391 Application crashes if download patient is opened and no inter…

hao555sky:
…net connection is present


hao555sky:
Sorry for errors I did several times.

PR #396, opened 2017-12-12
AC-426: Improve UI of login form

henziger:
Ref ticket: https://issues.openmrs.org/browse/AC-426

With this commit, we indicate that fields are required by adding a
"Required *" label at the bottom of the form's fields. Required fields
has a matching asterisk.

This commit uses HTML formatted strings to show
the user which fields in the login form are required
to be able to login. However, the formatting doesn't
work for TextInputLayout, only for TextView. Therefore,
we toggle between setting the hint on the TextView (to
get color coding) or on the TextInputLayout (to get
floating hints).

This commit also adds some styling to the login button:
Specifically, it turns it pale, not just the background
but the text color as well until all required fields are entered.

### DIFF
--- a/openmrs-client/src/main/res/values/strings.xml
+++ b/openmrs-client/src/main/res/values/strings.xml
@@ -27,6 +27,8 @@
     <string name="login_username_hint">Username</string>
     <string name="login_password_hint">Password</string>
     <string name="login_location_select">Session Location</string>
+    <string name="login_required">Required</string>
+    <string name="req_star"><![CDATA[<font color=\'#cc0029\'> * </font>]]></string>
     <string name="warning_dialog_title">Warning</string>
     <string name="warning_lost_data_dialog">Changing the URL or username will erase all the data. Are you sure?</string>
     <string name="forgot_dialog_title">Unable to login?</string>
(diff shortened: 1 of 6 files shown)


openmrs-client/src/main/res/values/strings.xml
```diff
@@ -26,7 +26,8 @@
     <string name="login_url_hint">URL address</string>
     <string name="login_username_hint">Username</string>
     <string name="login_password_hint">Password</string>
-    <string name="login_location_select">Session Location</string>
+    <string name="login_location_select"><![CDATA[Session Location <font color=\'#cc0029\'>*</font>]]></string>
```
DefCon-007: @henziger Adding color tags in strings.xml is not good. Change the color in activity only and keep only text here.

henziger: Thanks for the feedback! :)
I fixed it with the latest commit. I don't know why it's not good to have tags in strings.xml though.

shivtej1505: @DefCon-007, I think it's alright to use `CDATA` in this case because `*` will be same across all languages and official docs do support such kind of cases (see [here](https://developer.android.com/guide/topics/resources/string-resource.html)).

@henziger, sorry for late review. Could you define another string resource just with `*` with red color like this: `<string name="req_star">![CDATA[<font color='#cc0029'> *</font>]]</string>`. This way we aren't hardcoding the string and also reusing the new string.

Let me know if you have any question :)

henziger: Thanks @shivtej1505, and don't worry about being late :)

Good idea with separating the req_star string from the other strings, I've updated the PR.

shivtej1505: Thanks, @henziger. Merging now :)

PR #406, opened 2018-01-18
[AC-447] Add Data Validation when registering patients

f4ww4z:
https://issues.openmrs.org/browse/AC-447

Added a function in `ViewUtils` to validate Strings to check for illegal characters in text fields. Also added tests to the presenter.

Screenshots of the UI are [name_errors](https://issues.openmrs.org/secure/attachment/51120/name_errors.png) and [invalid_address](https://issues.openmrs.org/secure/attachment/51119/address_error.png).

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
@@ -226,6 +226,23 @@ public class AddEditPatientFragment extends ACBaseFragment<AddEditPatientContrac
     private Person createPerson() {
         Person person = new Person();
 
+        String emptyError = getString(R.string.emptyerror);
+
+        // Validate address
+        if (ViewUtils.isEmpty(edaddr1)
+                && ViewUtils.isEmpty(edaddr2)
+                && ViewUtils.isEmpty(edcity)
+                && ViewUtils.isEmpty(edpostal)
+                && ViewUtils.isEmpty(edcountry)
+                && ViewUtils.isEmpty(edstate)) {
+
+            addrerror.setText(R.string.atleastone);
+        } else if (!ViewUtils.validateText(ViewUtils.getInput(edaddr1), ViewUtils.ILLEGAL_ADDRESS_CHARACTERS)
+                || !ViewUtils.validateText(ViewUtils.getInput(edaddr2), ViewUtils.ILLEGAL_ADDRESS_CHARACTERS)) {
+
+            addrerror.setText(getString(R.string.addr_invalid_error));
+        }
+
         // Add address
         PersonAddress address = new PersonAddress();
         address.setAddress1(ViewUtils.getInput(edaddr1));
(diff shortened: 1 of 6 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
```diff
@@ -240,6 +257,36 @@ private Person createPerson() {
         addresses.add(address);
         person.setAddresses(addresses);
 
+        // Validate names
+        String unvalidatedError = getString(R.string.name_invalid_error);
```
shivtej1505: Give another name

openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
```diff
@@ -240,6 +257,36 @@ private Person createPerson() {
...
+        String middleNameError = getString(R.string.midname_invalid_error);
+        String givenMiddleNameError = getString(R.string.fmname_invalid_error);
```
shivtej1505: Can we clearly mention in which field contain invalid characters? Like here, we didn't include anything about `given name` and `family name` so that might be confusing for users.

openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
```diff
@@ -240,6 +257,36 @@ private Person createPerson() {
...
+
+        if (ViewUtils.isEmpty(edfname)) {
+            // Require first name
+            fnameerror.setText(emptyError);
+        } else if (!ViewUtils.validateText(ViewUtils.getInput(edfname), ViewUtils.ILLEGAL_CHARACTERS)) {
```
shivtej1505: We can put this line in separate function as it is being used in at least 3 places.

openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
```diff
@@ -240,6 +257,36 @@ private Person createPerson() {
...
+            fnameerror.setText(unvalidatedError);
+        }
+
+        // Middle name invalid error check if filled
+        if (!ViewUtils.isEmpty(edmname) &&
+                !ViewUtils.validateText(ViewUtils.getInput(edmname), ViewUtils.ILLEGAL_CHARACTERS)) {
+            // SHow invalid entry for both text
+            if (fnameerror.getText().toString().equals(unvalidatedError)) {
```
shivtej1505: We should omit this kind of logic, as this statement check the string which is there because of invalid `first name`. We should directly check if the `first name` field is invalid or not, we shouldn't check the consequence due to the invalid field.

shivtej1505: Also, you can first check for `middle name` and `first name` field together. It would be more readable and understandable logic.

openmrs-client/src/main/java/org/openmrs/mobile/utilities/ViewUtils.java
```diff
@@ -16,8 +16,14 @@
 
 import android.widget.EditText;
 
+import java.util.regex.Pattern;
+
 public class ViewUtils {
 
+    public static final String ILLEGAL_CHARACTERS = "[$&+,:;=\\\\?@#|/'<>.^*()%!-]";
```
shivtej1505: What about the names like `Jean-paul Sartre` and `John Q. Public`.
See here: https://www.google.co.in/search?q=jean-paul+sartre

f4ww4z: I just saw https://www.w3.org/International/questions/qa-personal-names and I think `,` and `-` should be allowed. @shivtej1505 Please tell me if I am missing characters which are also allowed in names.

f4ww4z:
@shivtej1505 I have made your required changes. I removed `,`, `.` and `-` from `ILLEGAL_CHARACTERS` and removed the `.` from `ADDRESS`. Also made the if statements easier to read.

PR #428, opened 2018-08-02
AC-467 : user details Change detection

csmuthukuda:
Currently, there is no any way to detect whether the user has been changed the user credentials from the web app. This issue will implement a new service to continuously check for user details while the application is active and log out the user if credentials have been changed.

Issue Link: https://issues.openmrs.org/browse/AC-467

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/services/AuthenticateCheckService.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/services/AuthenticateCheckService.java
@@ -1,0 +1,125 @@
+package org.openmrs.mobile.services;
+
+import android.app.Service;
+import android.content.Intent;
+import android.os.Binder;
+import android.os.IBinder;
+import android.util.Log;
+
+import org.openmrs.mobile.activities.login.LoginActivity;
+import org.openmrs.mobile.api.RestApi;
+import org.openmrs.mobile.api.RestServiceBuilder;
+import org.openmrs.mobile.application.OpenMRS;
+import org.openmrs.mobile.databases.OpenMRSDBOpenHelper;
+import org.openmrs.mobile.models.Session;
+import org.openmrs.mobile.utilities.ApplicationConstants;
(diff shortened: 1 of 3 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/services/AuthenticateCheckService.java
```diff
@@ -0,0 +1,125 @@
...
+                            Log.e("Service Task ", "user authenticated");
+
+                        } else {
+                            Log.e("Service Task ", "User Credentials Changed");
+                            OpenMRSDBOpenHelper.getInstance().closeDatabases();
+                            mOpenMRS.clearUserPreferencesData();
+                            mOpenMRS.clearCurrentLoggedInUserInfo();
+                            Intent intent = new Intent(getApplicationContext(), LoginActivity.class);
+                            intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+                            getApplicationContext().startActivity(intent);
```
f4ww4z: The user should be informed that the login credentials is incorrect anymore, before going to the Login screen. I think the user should be shown a dialog with positive and negative buttons.

dkayiwa:
@csmuthukuda are you still working on this?

f4ww4z:
@csmuthukuda Ping me if you want to resume working on it.

openmrs-client/src/main/java/org/openmrs/mobile/services/AuthenticateCheckService.java
VANKINEENITAWRUN: @f4ww4z can you please mention what would be the action of negative and positive buttons would be?

f4ww4z: @VANKINEENITAWRUN so when the user's changed credentials from the web app, the user should be shown a dialog first that shows the user's credentials changed, and that he will be logged out. I think that only 1 button (text is 'OK') is needed.

PR #437, opened 2018-11-14
AC-465: Encrypt database from the details derived from the username and password

anuar2k:
This PR brings several changes to the code
- User's password is not stored as a plaintext (used in offline verification), but as a bcrypt hash. It makes it not possible to steal user's password from `SharedPreferences` anymore.
- SQLite database uses a peppered bcrypt hash as its password, which is derived from user's name and password. It is not stored on the device anymore, making it impossible to steal, but generated after the user enters the password.
- When the user logs in to the same account using new password, the user is warned and the DB gets wiped (because the app just breaks when user changes the pass, doesn't matter if it's online or offline and because the new password is not sufficient for generating matching key to SQLite db). This is a temporary fix to lack of handling users who have changed their passwords. The problem with these users should be addressed in future tickets/PR's/whatever (e. g. user should be asked for old password to unlock the DB locked with old pass).
- The tests have been modified to comply with new code.

**All existing tests passed.**
## Issue I worked on

https://issues.openmrs.org/browse/AC-465

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/application/OpenMRS.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/application/OpenMRS.java
@@ -25,6 +25,7 @@ import com.activeandroid.Configuration;
 
 import net.sqlcipher.database.SQLiteDatabase;
 
+import org.mindrot.jbcrypt.BCrypt;
 import org.openmrs.mobile.api.FormListService;
 import org.openmrs.mobile.databases.OpenMRSDBOpenHelper;
 import org.openmrs.mobile.models.EncounterType;
@@ -32,7 +33,6 @@ import org.openmrs.mobile.models.Encountercreate;
 import org.openmrs.mobile.models.FormResource;
 import org.openmrs.mobile.models.Link;
 import org.openmrs.mobile.models.Obscreate;
-import org.openmrs.mobile.security.SecretKeyGenerator;
 import org.openmrs.mobile.utilities.ApplicationConstants;
 
 import java.io.File;
@@ -47,6 +47,7 @@ public class OpenMRS extends Application {
 
     private static OpenMRS instance;
     private OpenMRSLogger mLogger;
+    private String secretKey;
 
     @Override
     public void onCreate() {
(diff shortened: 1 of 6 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/application/OpenMRS.java
```diff
@@ -104,6 +105,21 @@ public void setPassword(String password) {
         editor.commit();
     }
 
+    public void setHashedPassword(String hashedPassword) {
+        SharedPreferences.Editor editor = getOpenMRSSharedPreferences().edit();
+        editor.putString(ApplicationConstants.UserKeys.HASHED_PASSWORD, hashedPassword);
+        editor.commit();
```
f4ww4z: It is best practice to use `editor.apply()` as `commit()` should not be called from the UI thread.

openmrs-client/src/main/java/org/openmrs/mobile/application/OpenMRS.java
```diff
@@ -104,6 +105,21 @@ public void setPassword(String password) {
...
+    }
+
+    public void setPasswordAndHashedPassword(String password) {
+        SharedPreferences.Editor editor = getOpenMRSSharedPreferences().edit();
+        String salt = BCrypt.gensalt(ApplicationConstants.DEFAULT_BCRYPT_ROUND);
+        String hashedPassword = BCrypt.hashpw(password, salt);
+        editor.putString(ApplicationConstants.UserKeys.PASSWORD, password);
+        editor.putString(ApplicationConstants.UserKeys.HASHED_PASSWORD, hashedPassword);
+        editor.commit();
```
f4ww4z: It is best practice to use `editor.apply()` as `commit()` should not be called from the UI thread.

openmrs-client/src/main/java/org/openmrs/mobile/application/OpenMRS.java
```diff
@@ -190,19 +211,20 @@ public String getVisitTypeUUID() {
...
-    private void generateKey() {
-        // create database key only if not exist
-        if (ApplicationConstants.EMPTY_STRING.equals(getSecretKey())) {
-            SharedPreferences.Editor editor = getOpenMRSSharedPreferences().edit();
-            String key = SecretKeyGenerator.generateKey();
-            editor.putString(ApplicationConstants.SECRET_KEY, key);
-            editor.commit();
-        }
+    private void createSecretKey() {
+        String inputPepper = "Open Sesame";
```
f4ww4z: This can possibly be put in `ApplicationConstants` as well.

PR #438, opened 2018-11-15
AC-483: Adding coveralls to openmrs-contrib-anroid-client repository

sviho:
I've added coveralls to the repository by changing build.gradle and travis.yml files.

## Issue I worked on
https://issues.openmrs.org/browse/AC-483


openmrs-client/build.gradle
```diff
@@ -1,8 +1,3 @@
-apply plugin: 'com.android.application'
-apply plugin: 'kotlin-android'
-apply plugin: 'kotlin-android-extensions'
-
-
```
f4ww4z: Is there a reason to move this block of code to the bottom? Is this the cause of the Travis CI failure?

f4ww4z:
Thanks for the PR @sviho ! Left a comment above.

openmrs-client/build.gradle
sviho: Yes, I think so. I've moved it because at first I've used plugins{} block and it should only go after buildscript{} otherwise Gradle fails. Since I use apply now, I've fixed and moved it to the top. Thanks!

sviho:
Reopened here https://github.com/openmrs/openmrs-contrib-android-client/pull/440

PR #512, opened 2019-03-26
AC-564: Annotate NotNull / Nullable where needed

TarekAlabd:
AC-564: Annotate NotNull / Nullable where needed

the issue I worked on:
https://issues.openmrs.org/browse/AC-564


TarekAlabd:
@deepak140596 I want to close that request and create a new one with the last commit only
how to do that ?

PR #542, opened 2019-04-14
AC-498 Added Delete Image for Uploaded image

VANKINEENITAWRUN:
Added an option to camera Dialog for Image Removal. As the Remove Photo option clicked change the Patientphoto to person_grey_icon Bitmap.

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-498

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientFragment.java
@@ -605,26 +605,31 @@ public class AddEditPatientFragment extends ACBaseFragment<AddEditPatientContrac
         capturePhotoBtn.setOnClickListener(view -> {
 
             CameraOrGalleryPickerDialog dialog = CameraOrGalleryPickerDialog.getInstance(
-
                     (dialog1, which) -> {
+                                if (which == 0) {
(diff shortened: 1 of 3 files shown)


codepoet2017390:
Also i think it would be better if the patient image is deleted from the server side as well. Present fix just deletes the image on the users device but the patient image still persists in the database leading to inconsistency in the implementers database. As far as i know there is no speicific API for deleting an image but this can be dealt with using the existing update APIs and updating the image with a generic placeholder (as already existing by defalut).

VANKINEENITAWRUN:
> Also i think it would be better if the patient image is deleted from the server side as well. Present fix just deletes the image on the users device but the patient image still persists in the database leading to inconsistency in the implementers database. As far as i know there is no speicific API for deleting an image but this can be dealt with using the existing update APIs and updating the image with a generic placeholder (as already existing by defalut).

VANKINEENITAWRUN:
> but this can be dealt with using the existing update APIs and updating the image with a generic placeholder (as already existing by default).

Exactly that's what is done. when a user removes the image, a bitmap of `R.drawable.ic_person_grey_500_48dp.png` gets updated as a patient profile. so the previous profile gets overwritten.

codepoet2017390:
Then just sync patient data.

codepoet2017390:
Ok just saw this is a part of the update patient activity. Looks good

f4ww4z:
@VANKINEENITAWRUN If you fix the merge conflict, I'm happy to merge this.

PR #557, opened 2019-04-20
AC587: Obsolete SDK_INT Version Check

vansha10:

Removed Android SDK_INT Version check where version to be checked was less than 16 (11 in this case), as the Minimum SDK Version for this project is 16.
Also removed the function that checks if API level is greater than equal to 16, which it always would be.

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-587

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/logs/LogsFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/logs/LogsFragment.java
@@ -63,16 +63,10 @@ public class LogsFragment extends ACBaseFragment<LogsContract.Presenter> impleme
         });
     }
 
-    @SuppressWarnings("deprecation")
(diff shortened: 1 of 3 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/activities/logs/LogsFragment.java
```diff
@@ -65,14 +65,9 @@ public void onClick(View v) {
 
     @SuppressWarnings("deprecation")
```
VANKINEENITAWRUN: @vansha10 suppress line can be deleted.

vansha10: @VANKINEENITAWRUN 
Thanks for the suggestions.

PR #586, opened 2019-05-26
AC-480 Patient model should extend Person

juniorjainsahab:
* Changing the relationship between `Patient` and `Person` to `is a` relationship instead of `has a` relationship.
* Details that can be updated in a patient now can be updated through a new `PatientDetails` object.
* We had a hard dependency on the backend API while updating or registering patient, It still accepts certain detail as part of `person` key. So we have exposed a `getPerson` method while making any network request.  

## Issue I worked on
After this PR, `Patient` model is extending Person. From now onwards, we have to get patient details from the Patient itself instead of from Person.
## Before
     patient.getPerson().getNames();
## After
     patient.getNames();

I have already run the unit test at my local machine and all are passing successfully and I also run the application on the device and it also works the expected way.

JIRA Issue: https://issues.openmrs.org/browse/AC-480

- [X] My pull request only contains **ONE single commit**

- [X] All new and existing **tests passed**.

- [X] My pull request is **based on the latest changes** of the master branch.

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientPresenter.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientPresenter.java
@@ -14,6 +14,8 @@
 
 package org.openmrs.mobile.activities.addeditpatient;
 
+import androidx.annotation.NonNull;
+
 import org.openmrs.mobile.activities.BasePresenter;
 import org.openmrs.mobile.api.RestApi;
 import org.openmrs.mobile.api.RestServiceBuilder;
@@ -34,7 +36,6 @@ import org.openmrs.mobile.utilities.ViewUtils;
 
 import java.util.List;
 
-import androidx.annotation.NonNull;
 import retrofit2.Call;
 import retrofit2.Callback;
 import retrofit2.Response;
@@ -128,7 +129,7 @@ public class AddEditPatientPresenter extends BasePresenter implements AddEditPat
         mPatientInfoView.setErrorsVisibility(givenNameError, familyNameError, dateOfBirthError, genderError, addressError, countryError, countryNull, stateError, cityError, postalError);
 
         // Validate names
-        PersonName currentPersonName = patient.getPerson().getName();
+        PersonName currentPersonName = patient.getName();
 
         if (StringUtils.isBlank(currentPersonName.getGivenName())
                 || !ViewUtils.validateText(currentPersonName.getGivenName(), ViewUtils.ILLEGAL_CHARACTERS)) {
@@ -146,13 +147,13 @@ public class AddEditPatientPresenter extends BasePresenter implements AddEditPat
         }
 
         // Validate date of birth
-        if (StringUtils.isBlank(patient.getPerson().getBirthdate())) {
+        if (StringUtils.isBlank(patient.getBirthdate())) {
             dateOfBirthError = true;
         }
 
         // Validate address
-        String patientAddress1 = patient.getPerson().getAddress().getAddress1();
-        String patientAddress2 = patient.getPerson().getAddress().getAddress2();
+        String patientAddress1 = patient.getAddress().getAddress1();
+        String patientAddress2 = patient.getAddress().getAddress2();
 
         if ((StringUtils.isBlank(patientAddress1)
                 && StringUtils.isBlank(patientAddress2)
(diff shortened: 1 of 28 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/activities/addeditpatient/AddEditPatientPresenter.java
```diff
@@ -20,8 +20,11 @@
 import org.openmrs.mobile.api.retrofit.PatientApi;
 import org.openmrs.mobile.dao.PatientDAO;
 import org.openmrs.mobile.listeners.retrofit.DefaultResponseCallbackListener;
+import org.openmrs.mobile.models.IdentifierType;
 import org.openmrs.mobile.models.Module;
 import org.openmrs.mobile.models.Patient;
+import org.openmrs.mobile.models.PatientDto;
+import org.openmrs.mobile.models.Person;
```
f4ww4z: @juniorjainsahab Please remove all unused imports.

openmrs-client/src/main/java/org/openmrs/mobile/models/Patient.java
```diff
@@ -13,6 +13,7 @@
 import com.google.gson.annotations.Expose;
 import com.google.gson.annotations.SerializedName;
 
+import org.openmrs.mobile.dao.PatientDAO;
```
f4ww4z: We shouldn't need a DAO import in a model class.

juniorjainsahab:
Creating setters for multiple fields in a class can be dangerous sometimes, One may forget to set some fields which will result in the `run time` failure. These things can be easily resolved at the language level on the compile time by setting up the `data contract` which a class needs to be initialized with. `PatientDetails` is one such class.

Regarding `PatientDto`, this class is something which is defining the structure with backend APIs, If i remove it and use `Patient` instead. backend server will send `400 bad request` since the structure of request body has changed. 

Let me know if you think otherwise and have a suggestion on those two points. Meanwhile, I'll be fixing the unused imports issue.

f4ww4z:
@juniorjainsahab What you can do is make a constructor in `Patient`, as currently the way to create a Patient object is create an empty `Patient` object and calling multiple setters. See [an example ](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Patient.java#L54)in the openmrs-core repo. You can edit the constructor to take in the same variables as in the `PatientDetails` class, so creating a `Patient` object with arguments defined in the constructor is the same as creating an empty object and calling setters on it. You can also overload the constructor, if necessary.

I think that `PatientDto` looks good to me, just add a javadoc in the class and describe how the class is used.

juniorjainsahab:
@f4ww4z all the changes are done.

juniorjainsahab:
@f4ww4z 
if you have some feedback do let me know otherwise please merge it as soon as possible because it will be hard to remove merge conflicts.

openmrs-client/src/main/java/org/openmrs/mobile/models/PatientDto.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+
+    /**
+     *
+     * @return
+     *     The patient
+     */
+    public Patient getPatient(){
+        Patient patient = new Patient();
```
f4ww4z: @juniorjainsahab I requested to make a `Patient` constructor, so you have not done that part. The constructor is very important so that in the future we do not forget to set some attributes. It also makes the code more concise.

openmrs-client/src/main/java/org/openmrs/mobile/models/Person.java
```diff
@@ -45,9 +45,7 @@
     private Bitmap photo;
 
     /**
-     * 
-     * @return
-     *     The names
+     * @return The names
```
f4ww4z: This doesn't improve the documentation. Revert the changes to this **whole file** or properly describe the variables.

openmrs-client/src/main/java/org/openmrs/mobile/models/PatientDto.java
```diff
@@ -0,0 +1,81 @@
+package org.openmrs.mobile.models;
```
f4ww4z: Add the OpenMRS license header at the start.

openmrs-client/src/main/java/org/openmrs/mobile/models/PatientDto.java
```diff
@@ -0,0 +1,81 @@
...
+
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PatientDto {
```
f4ww4z: I mentioned describing this class with a javadoc.

openmrs-client/src/main/java/org/openmrs/mobile/models/PatientDto.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @SerializedName("person")
+    @Expose
+    private Person person;
+
+    /**
+     *
+     * @return
+     *     The uuid
+     */
```
f4ww4z: This doesn't improve the documentation. You can either describe the variable or remove the javadoc. Do this for all the method docs in this file.

openmrs-client/src/main/java/org/openmrs/mobile/models/PatientDto.java
juniorjainsahab: this looks like a really big change since `AddEditPatientFragment` holds one of the patient object. and the very same object keeps updating every time when the user clicks a submit button.

For an issue like this where we just had to extend Patient with Person and establish an `is a` relationship. these kinds of change may totally change the app's behavior and can cause the app to crash and that's why application again needs to be thoroughly tested. We have already change the create/update API's payload to a different object `PatientDto` and that works in an expected manner.

I'd recommend to not to create constructor this time and fix it in an iterative manner. Could you please create a new issue for that, I'd be happy to pick that up and fix all mutability issue.

Let me know if you think otherwise.

juniorjainsahab:
@f4ww4z , I think there has been some change regarding play store stuff and causing both of the build services to fail. As far as I know, this is not related to my changes. Could you please verify that?

openmrs-client/src/main/java/org/openmrs/mobile/models/PatientDto.java
f4ww4z: I now realize that this change will be bigger than expected, but still an issue that needs fixing. @juniorjainsahab do file a new issue for this, since I may forget.

juniorjainsahab:
@f4ww4z I have squashed all the commits into one. Please let me know if something is required from my side.

openmrs-client/src/main/java/org/openmrs/mobile/models/PatientDto.java
juniorjainsahab: @f4ww4z new issue is also created you can check it on [AC-617](https://issues.openmrs.org/browse/AC-617).

f4ww4z:
@juniorjainsahab The external CI error is now fixed - please pull the latest changes from master.

juniorjainsahab:
@f4ww4z now both the ci builds are passing. So it is now ready for merge.

PR #605, opened 2019-06-06
AC-614: Strange error occurs while Testing

ODORA0:
**Issue worked on**: https://issues.openmrs.org/browse/AC-614

**Work Done**: Deleted `@PrepareForTest({ Cache.class, TableInfo.class, Context.class,
        ContentResolver.class, ContentProvider.class, ContentValues.class })` and

` @SuppressStaticInitializationFor("com.activeandroid.content.ContentProvider")`

### DIFF
--- a/openmrs-client/src/test/java/org/openmrs/mobile/test/ACUnitTestBase.java
+++ b/openmrs-client/src/test/java/org/openmrs/mobile/test/ACUnitTestBase.java
@@ -51,10 +51,7 @@ import static org.mockito.Matchers.anyObject;
 import static org.mockito.Mockito.doNothing;
 import static org.powermock.api.mockito.PowerMockito.when;
 
-@PrepareForTest({ Cache.class, TableInfo.class, Context.class,
-        ContentResolver.class, ContentProvider.class, ContentValues.class })
 @RunWith(PowerMockRunner.class)
-@SuppressStaticInitializationFor("com.activeandroid.content.ContentProvider")
 public abstract class ACUnitTestBase {
 
     @Before


deepak140596:
@ODORA0 , looking at the Build failure, I think that I was wrong. 

I was looking into this matter and found out that there are 2 classes `PatientDashboardDiagnosisPresenterTest` and `FormListPresenterTest` that are causing the tests to fail on Android Studio. 

Removing those lines enables Test to run on Android Studio with 8 test failing and rest passing. The main reason is the `Cache` class. Also after hours of debugging, I found out that we shouldn't remove those lines. Instead we need to replace/simulate `Cache` class in some other way.

ODORA0:
Yes!! So what are you suggesting?

deepak140596:
@ODORA0 , my suggestion will be that you can debug the cause of error and try to replace the error causing code with alternative code. If the test runs successfully on your system, then again refresh this pull request to see if they are running well on the TravisCI.

It will take some amount of effort but the good thing is that only the above  Test classes `PatientDashboardDiagnosisPresenterTest` and `FormListPresenterTest` are just causing the problem. Rest of the classes are working fine.

f4ww4z:
@ODORA0 Did you get a chance to debug further?

HerbertYiga:
according to your logs,am seeing some deprecated gradle feature which is incompatible with gradle 6.0,try to look through your logs vigilantly

deepak140596:
@HerbertYiga , can you show us some logs message. I am having trouble finding them.

HerbertYiga:
deepak140596 could you check on details of continuous-intergration on this page!!

ODORA0:
Yes @HerbertYiga I see that in line 897 @deepak140596, although the problem comes when for every change I make, freezes my PC I don't really get that

f4ww4z:
@ODORA0 based on your comment on JIRA, you shouldn't be using maven at all to test the app. Just simply right click the `test` folder and run all tests.

Also, see [this error line](https://travis-ci.org/openmrs/openmrs-contrib-android-client/builds/542379666#L852). Check `DateUtilsTest` for any deprecated features.

f4ww4z:
@ODORA0 Any updates?

PR #633, opened 2019-12-13
AC-530 The form edit button in Vitals page is not working

PermissionError:

Upon investigation, I have discovered that when signing in to OpenMRS AC for the first time, `FormListService` is not called due to the user not being logged in. After the user authenticates successfully, they cannot use any forms, since they have not been downloaded. I have fixed this by calling `FormListService` upon user authentication in `LoginFragment`.

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-530

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginFragment.java
@@ -359,6 +359,8 @@ public class LoginFragment extends ACBaseFragment<LoginContract.Presenter> imple
         Intent intent = new Intent(mOpenMRS.getApplicationContext(), DashboardActivity.class);
         intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
         mOpenMRS.getApplicationContext().startActivity(intent);
+        Intent formListServiceIntent = new Intent(mOpenMRS.getApplicationContext(), FormListService.class);
+        mOpenMRS.getApplicationContext().startService(formListServiceIntent);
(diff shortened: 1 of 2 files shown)


PermissionError:
> @PermissionError can you see the decretion of the test coverage here.

Thank you for the pointers @Akayeshmantha, but there doesn't seem to be any existing tests for UI fragments that I can take an example of. Does an entire suite of tests need to be written for LoginFragment?

openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginFragment.java
```diff
@@ -359,6 +359,8 @@ public void userAuthenticated() {
         Intent intent = new Intent(mOpenMRS.getApplicationContext(), DashboardActivity.class);
         intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
         mOpenMRS.getApplicationContext().startActivity(intent);
+        Intent intent1 = new Intent(mOpenMRS.getApplicationContext(), FormListService.class);
```
f4ww4z: Change `intent1` to a more meaningful name, e.g. `formServiceIntent`

PermissionError:
Thanks for the tips! A new commit has been force-pushed.

The use of a try-catch block is due to the fact that the method would usually throw a NullPointerException, since there is no mDropdownList, but using JUnit's expected exception still ends the test execution prematurely. Thus, a good old try-catch was used.

The verify() testing for calling the method exactly once. I have attempted to check whether the submitted Intent has a class of FormListService, but was unsuccessful despite following the android documentations. I have even used two identical Intents, i.e. `new Intent(context, FormListService.class)`, but Intent#filterEquals still returns false. So I had to resolve to a compromise.

PR #645, opened 2019-12-24
AC-635:Unit Test Error Fix

ishaan-das:

The class that was causing a problem was Cache.java which was a child of ActiveAndroid and was using android.support.v4.LruCache instead of android.util.LruCache which is the currently supported version. But this couldn't be changed directly in the Cache.java class as it was read only, so I created another class new_cache.java and transferred everything to it to fix the Unit Test Issue.
I also created new_ModelInfo.java class because the original one which was a child of ActiveAndroid couldn't be imported and used this new one in new_cache.java

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-635

### DIFF
--- a/openmrs-client/src/test/java/org/openmrs/mobile/test/new_ModelInfo.java
+++ b/openmrs-client/src/test/java/org/openmrs/mobile/test/new_ModelInfo.java
@@ -1,0 +1,193 @@
+package org.openmrs.mobile.test;
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.Collection;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import android.content.Context;
+import com.activeandroid.Configuration;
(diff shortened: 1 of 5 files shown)


openmrs-client/src/test/java/org/openmrs/mobile/test/new_ModelInfo.java
```diff
@@ -0,0 +1,193 @@
...
+            }
+            catch (InstantiationException e) {
+                Log.e("Couldn't instantiate TypeSerializer.", e);
+            }
+            catch (IllegalAccessException e) {
+                Log.e("IllegalAccessException", e);
+            }
+        }
+    }
+}
```
f4ww4z: Reformat this file with Android Studio's default formatter.

ishaan-das: hi sir,
I have implemented all the suggested changes and have created a new pull request. Kindly review it.

ribhavsharma:
@ishaan-das I see you've created an updated PR, so close this one maybe?

PR #698, opened 2020-02-09
AC-684 - Show 'yes' or 'no' instead of random characters in active visit details

sanjulamadurapperuma:

I added validations to make sure that the HIV status is shown to the user when viewing the Visit Notes if available.

image

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-684

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitExpandableListAdapter.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitExpandableListAdapter.java
@@ -79,8 +79,19 @@ public class VisitExpandableListAdapter extends BaseExpandableListAdapter {
                             convertView = openMRSInflater.addKeyValueStringView(contentLayout, obs.getDiagnosisOrder(),
                                     "(" + obs.getShortDiagnosisCertainty() + ") " + obs.getDiagnosisList());
                         } else if (obs.getDisplay() != null && obs.getDisplayValue() != null) {
-                            //miscellaneous, for all other cases that have a Display - Value pair
-                            convertView = openMRSInflater.addKeyValueStringView(contentLayout, obs.getDisplay(), obs.getDisplayValue());
+                            if (obs.getDisplay().contains(mContext.getString(R.string.hiv_yes))) {
+                                convertView = openMRSInflater.addKeyValueStringView(contentLayout, obs.getDisplay(),
+                                        mContext.getString(R.string.hiv_yes));
+                            } else if (obs.getDisplay().contains(mContext.getString(R.string.hiv_no))) {
+                                convertView = openMRSInflater.addKeyValueStringView(contentLayout, obs.getDisplay(),
+                                        mContext.getString(R.string.hiv_no));
+                            } else if (obs.getDisplay().contains(mContext.getString(R.string.hiv_unknown))) {
+                                convertView = openMRSInflater.addKeyValueStringView(contentLayout, obs.getDisplay(),
+                                        mContext.getString(R.string.hiv_unknown));
+                            } else {
+                                //miscellaneous, for all other cases that have a Display - Value pair
+                                convertView = openMRSInflater.addKeyValueStringView(contentLayout, obs.getDisplay(), obs.getDisplayValue());
+                            }
                         }
                     }
                     layouts.add(convertView);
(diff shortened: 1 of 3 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitExpandableListAdapter.java
```diff
@@ -40,6 +40,9 @@
 
     private static final int LEFT = 0;
     private static final int RIGHT = 1;
+    private static final String HIV_YES = "Yes";
+    private static final String HIV_NO = "No";
+    private static final String HIV_UNKNOWN = "Unknown";
```
LuGO0: instead of using static constants it would be better if we used strings.xml file for referencing string constants since these string will be displayed to the user and hence it would also help with localization of these strings.

AryanGanotra07:
Hi @sanjulamadurapperuma you pull request checkboxes have some format issue. Please correct it.

openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitExpandableListAdapter.java
f4ww4z: @sanjulamadurapperuma Did you see this comment? It's valid.

openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitExpandableListAdapter.java
```diff
@@ -79,8 +82,16 @@ public VisitExpandableListAdapter(Context context, List<Encounter> encounters) {
                             convertView = openMRSInflater.addKeyValueStringView(contentLayout, obs.getDiagnosisOrder(),
                                     "(" + obs.getShortDiagnosisCertainty() + ") " + obs.getDiagnosisList());
                         } else if (obs.getDisplay() != null && obs.getDisplayValue() != null) {
-                            //miscellaneous, for all other cases that have a Display - Value pair
-                            convertView = openMRSInflater.addKeyValueStringView(contentLayout, obs.getDisplay(), obs.getDisplayValue());
+                            if (obs.getDisplay().contains(HIV_YES)) {
```
f4ww4z: Use switch case instead, it's more concise.

openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitExpandableListAdapter.java
sanjulamadurapperuma: Just saw this, I must have missed it. @LuGO0 @f4ww4z I'll work on this

sanjulamadurapperuma:
> Hi @sanjulamadurapperuma you pull request checkboxes have some format issue. Please correct it.

Sure will do, thanks

openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitExpandableListAdapter.java
sanjulamadurapperuma: Could we use a switch case here in this scenario? The objective is to check whether the string returned by obs.getDisplay() contains either one of the three values, Yes, No or Unknown, each of which could be turned into cases. However, the condition for the switch would require if condition before it, which would make the conversion from if else to switch case ineffective. What do you think @f4ww4z ?

openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitExpandableListAdapter.java
sanjulamadurapperuma: Changes made, thanks for pointing out :)

sanjulamadurapperuma:
> Hi @sanjulamadurapperuma . Please squash all commits.

After merging, my commits (spread over several days) have dispersed, so I won't be able to squash them :(

rishabh-997:
> > Hi @sanjulamadurapperuma . Please squash all commits.
> 
> After merging, my commits (spread over several days) have dispersed, so I won't be able to squash them :(

try using
`git pull --rebase upstream master`
this will collect all your commits together and then you can squash then

sanjulamadurapperuma:
> > > Hi @sanjulamadurapperuma . Please squash all commits.
> > 
> > 
> > After merging, my commits (spread over several days) have dispersed, so I won't be able to squash them :(
> 
> try using
> `git pull --rebase upstream master`
> this will collect all your commits together and then you can squash then

Oh! Silly me, yes this should work! Must have overlooked it, thanks :)

PR #714, opened 2020-03-22
AC - 723 Illegal State Exception - LoginFragment not attached to context

AryanGanotra07:

There was crash due to referencing the parent activity after the activity was destroyed.
Added not null checks to prevent crashes.

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-723

- [ ] My pull request only contains **ONE single commit**

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginFragment.java
@@ -329,19 +329,21 @@ public class LoginFragment extends ACBaseFragment<LoginContract.Presenter> imple
         mLastCorrectURL = serverURL;
         mUrl.setText(serverURL);
         mLocationsList = locationsList;
-        List<String> items = getLocationStringList(locationsList);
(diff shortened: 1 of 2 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/activities/login/LoginFragment.java
```diff
@@ -431,7 +435,8 @@ public void hideURLDialog() {
         if (mLocationsList == null) {
             mPresenter.loadLocations(mLastCorrectURL);
         } else {
-            initLoginForm(mLocationsList, mLastCorrectURL);
+            if (isActivityNotNull())
```
rishabh-997: was this necessary since we are already checking conditions inside `initLoginForm` ???

AryanGanotra07: It will not execute the function initLoginForm if activity is null, and if not null checks in the function at getLocationStringList.

HerbertYiga:
could you squash your commits!!

PR #735, opened 2020-05-12
AC-755 Fixed LocationRoomDAO tests

rishabh-997:
## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-755

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/dao/LocationRoomDAO.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/dao/LocationRoomDAO.java
@@ -26,7 +26,7 @@ import org.openmrs.mobile.databases.entities.LocationEntity;
 public interface LocationRoomDAO {
 
     @Insert
-    Single<Long> saveLocation(LocationEntity entity);
+    long saveLocation(LocationEntity entity);
 
     @Query("DELETE FROM locations")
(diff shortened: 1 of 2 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/dao/LocationRoomDAO.java
```diff
@@ -26,7 +26,7 @@
 public interface LocationRoomDAO {
 
     @Insert
-    Single<Long> saveLocation(LocationEntity entity);
+    void saveLocation(LocationEntity entity);
```
f4ww4z, line 29: Shouldn't this method return the new inserted location id? What is the change here?

rishabh-997: Using Single<Long> wasn't adding files to the database so I changed it to void and worked fine. I thought of writing other tests and look into it later but it completely slipped off my mind.

rishabh-997: anyways, fixed it... I have made return type as `Long` which is the same as other `entities`

VibhorChinda:
@rishabh-997 just squash commits. Everything else looks fine to me :)

PR #776, opened 2020-07-06
AC-478 Encrypt database with SQLCipher library

rishabh-997:
The` cwac-saferoom` library is now deprecated and hasn't been into development lately as the sql-cipher:4.4.0 provides support for the room encryption, so I read about it and tried implementing it

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-478

### DIFF
--- a/openmrs-client/build.gradle
+++ b/openmrs-client/build.gradle
@@ -144,6 +144,7 @@ dependencies {
     //noinspection GradleDependency
     annotationProcessor "androidx.room:room-compiler:${versions.roomCompiler}"
     implementation "net.zetetic:android-database-sqlcipher:${versions.sqlcipher}"
+    implementation "androidx.sqlite:sqlite:${versions.sqlAndroidx}"
 
     // Work Manager
     implementation "androidx.work:work-runtime:${versions.workManager}"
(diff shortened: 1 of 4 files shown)


rishabh-997:
Here is the link to implementation:
https://github.com/sqlcipher/android-database-sqlcipher#using-sqlcipher-for-android-with-room
After accessing the provider dashboard, the previous database still gets corrupted which I am not getting why. Is there any way to check if the provider database is getting encrypted or not.

LuGO0:
@rishabh-997 there is a lot of confusion regarding which model are being used and which arent I faced this while working on the provider module. maybe the problem might be there I will read it and then say !!

LuGO0:
@rishabh-997 maybe its because we arent using  all the annotated entities in the AppDatabase.java # like , the PatientEntity we arent using that.
Thats why I think we will first have to migrate completely from tables to room entities and then implement sqlCipher

f4ww4z:
>  After accessing the provider dashboard, the previous database still gets corrupted which I am not getting why. Is there any way to check if the provider database is getting encrypted or not.

Since the app is still using the activeandroid library for local database (and not using Room), we cannot check for that yet. We have to migrate all entities to the respective Room classes, e.g. `Person` to `PersonEntity`. I think this PR will be open until all migrations are complete, like discussed in the JIRA issue.

rishabh-997:
> > After accessing the provider dashboard, the previous database still gets corrupted which I am not getting why. Is there any way to check if the provider database is getting encrypted or not.
> 
> Since the app is still using the activeandroid library for local database (and not using Room), we cannot check for that yet. We have to migrate all entities to the respective Room classes, e.g. `Person` to `PersonEntity`. I think this PR will be open until all migrations are complete, like discussed in the JIRA issue.

Can you create a sample issue as subtask in jira explaining how exactly I need to change person to person entity... I will add others afterwards...

f4ww4z:
@rishabh-997 How is your progress for this?

rishabh-997:
> How is your progress for this

I will need to do some research on encrypting the existing database also using SQLCipher+Saferoom by using only a few snippets from saferoom rather than adding the whole dependency so this PR is redundant as I will have to start fresh... I will be closing this one and work on a new one...

PR #782, opened 2020-07-15
AC-813 added elided repository layer in visitDashboard package

LuGO0:
1. Added the missing repository layer to the visitDashboard package where the API endpoints were directly being accessed form the presenter instead of VisitsRepository .
2. fixed broken tests due to the change

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-813

- [ ] My pull request only contains **ONE single commit**

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitDashboardPresenter.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitDashboardPresenter.java
@@ -47,12 +47,20 @@ public class VisitDashboardPresenter extends BasePresenter implements VisitDashb
         visitRepository = new VisitRepository();
     }
 
+    /**
+     * Mock presenter used in the unit Tests
+     * @param restApi
+     * @param visitDAO
+     * @param visitId
+     * @param mVisitDashboardView
+     */
     public VisitDashboardPresenter(RestApi restApi, VisitDAO visitDAO, Long visitId, VisitDashboardContract.View mVisitDashboardView) {
         this.mVisitDashboardView = mVisitDashboardView;
         this.visitDAO = visitDAO;
         this.visitId = visitId;
         this.restApi = restApi;
         mVisitDashboardView.setPresenter(this);
+        visitRepository = new VisitRepository(restApi,visitDAO,null,null);
     }
 
     public void endVisitByUUID(final Visit visit) {
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitDashboardPresenter.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitDashboardPresenter.java
@@ -14,13 +14,13 @@
 
 package org.openmrs.mobile.activities.visitdashboard;
 
-import androidx.annotation.NonNull;
-
 import org.openmrs.mobile.R;
 import org.openmrs.mobile.activities.BasePresenter;
 import org.openmrs.mobile.api.RestApi;
 import org.openmrs.mobile.api.RestServiceBuilder;
+import org.openmrs.mobile.api.repository.VisitRepository;
 import org.openmrs.mobile.dao.VisitDAO;
+import org.openmrs.mobile.listeners.retrofit.DefaultVisitsCallback;
 import org.openmrs.mobile.models.Encounter;
 import org.openmrs.mobile.models.Patient;
 import org.openmrs.mobile.models.Visit;
@@ -29,9 +29,6 @@ import org.openmrs.mobile.utilities.StringUtils;
 
 import java.util.List;
 
-import retrofit2.Call;
-import retrofit2.Callback;
-import retrofit2.Response;
 import rx.android.schedulers.AndroidSchedulers;
 
 public class VisitDashboardPresenter extends BasePresenter implements VisitDashboardContract.Presenter {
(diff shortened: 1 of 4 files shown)


LuGO0:
@rishabh-997 @f4ww4z Sir please review !!

LuGO0:
and I also need some help in the way I should ensure that the callbacks are a cohesive unit all together and we dont end up making a new callbackListener interface for each method, what should I do how should an interface be designed so that It is generic and can be used in a wide number of functions ??

openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitDashboardPresenter.java
```diff
@@ -62,28 +69,22 @@ public void endVisitByUUID(final Visit visit) {
...
-                                .observeOn(AndroidSchedulers.mainThread())
-                                .subscribe(id -> moveToPatientDashboard());
-                        }));
-                } else {
-                    mVisitDashboardView.showErrorToast(response.message());
-                }
+            public void onSuccess(String response) {
+                addSubscription(visitDAO.getVisitByID(visit.getId())
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe(vis -> {
```
rishabh-997, line 77: rename `vis` to `visit`

openmrs-client/src/main/java/org/openmrs/mobile/activities/visitdashboard/VisitDashboardPresenter.java
```diff
@@ -29,31 +29,38 @@
...
+     * @param visitId
+     * @param mVisitDashboardView
+     */
     public VisitDashboardPresenter(RestApi restApi, VisitDAO visitDAO, Long visitId, VisitDashboardContract.View mVisitDashboardView) {
         this.mVisitDashboardView = mVisitDashboardView;
         this.visitDAO = visitDAO;
         this.visitId = visitId;
         this.restApi = restApi;
         mVisitDashboardView.setPresenter(this);
+        visitRepository = new VisitRepository(restApi,visitDAO,null,null);
```
rishabh-997, line 63: follow java code style conventions. add a space after the comma.

openmrs-client/src/main/java/org/openmrs/mobile/api/repository/VisitRepository.java
```diff
@@ -151,6 +160,30 @@ public void onFailure(@NonNull Call<Results<Encounter>> call, @NonNull Throwable
         });
     }
 
+    public void endVisitByUuid(String uuid, Visit test, DefaultVisitsCallback callbackListener) {
```
rishabh-997, line 163: rename `test` to `visit`

openmrs-client/src/main/java/org/openmrs/mobile/listeners/retrofit/DefaultVisitsCallback.java
```diff
@@ -0,0 +1,7 @@
+package org.openmrs.mobile.listeners.retrofit;
```
rishabh-997, line 1: Add openmrs copyright at the top

rishabh-997:
> We don't end up making a new callback listener interface for each method, what should I do how should an interface be designed so that It is generic and can be used in a wide number of functions ??

We can use templates to return the different objects as per response collected. But as of now, I think this should be kept on hold as these changes will clash with the migration task. I will look for how to create a generic listener meanwhile.

LuGO0:
> > We don't end up making a new callback listener interface for each method, what should I do how should an interface be designed so that It is generic and can be used in a wide number of functions ??
> 
> We can use templates to return the different objects as per response collected. But as of now, I think this should be kept on hold as these changes will clash with the migration task. I will look for how to create a generic listener meanwhile.

yes @rishabh-997 I will also leave it for now and do refactorings with new changes as callbacks which can later be refactored when we make generic callbacks working after room migration 👍

LuGO0:
> We can use templates to return the different objects as per response collected. But as of now, I think this should be kept on hold as these changes will clash with the migration task. I will look for how to create a generic listener meanwhile.

I looked for it as of now I think we will have to use wildcards <T extends someInterface> of the desired response we want on the presenter to be returned form the repository.

openmrs-client/src/main/java/org/openmrs/mobile/api/repository/VisitRepository.java
```diff
@@ -151,6 +160,30 @@ public void onFailure(@NonNull Call<Results<Encounter>> call, @NonNull Throwable
         });
     }
 
+    public void endVisitByUuid(String uuid, Visit visit, DefaultVisitsCallback callbackListener) {
+        restApi.endVisitByUUID(uuid, visit).enqueue(new Callback<Visit>() {
+            @Override
+            public void onResponse(@NonNull Call<Visit> call, @NonNull Response<Visit> response) {
+                if (response.isSuccessful()) {
+                    if (callbackListener != null) {
```
f4ww4z, line 168: Move this condition up, the `response.isSuccessful()` condition should be the one that's nested. That way you can check if the callbackListener is null and return an appropriate error message.

LuGO0: @f4ww4z can you please look up here https://github.com/openmrs/openmrs-contrib-android-client/pull/782/files#diff-c9d7275119a4fedc35414396ff152b8fR146 if you still think that this change should be made then I would do that 👍 .

f4ww4z: @LuGO0 Actually can you also make a similar change to the if statements on line 146, it was hidden while reviewing. This will reduce a reader's cognitive load.

LuGO0: okay @f4ww4z Sir I would do the same 👍

LuGO0: @f4ww4z sir made the requested change , but It would not work on line 146 though since there are a few things that need to be done on other than the callback listener there .

PR #789, opened 2020-07-30
AC-823 Overloaded methods of AppDataBase Helper class

rishabh-997:
There are 2 types of converters: convert a resource model class to the entity, and from entity to resource. We can group all these converters by converting the file to Kotlin first, then make the method name the same, just with different parameters.

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-823


openmrs-client/src/main/java/org/openmrs/mobile/databases/AppDatabaseHelper.kt
```diff
@@ -0,0 +1,315 @@
...
+            allergyEntity.severityDisplay = allergy.severity!!.display
+            allergyEntity.severityUUID = allergy.severity!!.uuid
+        }
+        allergyEntity.allergenDisplay = allergy.allergen!!.codedAllergen!!.display
+        allergyEntity.allergenUUID = allergy.allergen!!.codedAllergen!!.uuid
+        allergyEntity.allergyReactions = allergy.reactions
+        return allergyEntity
+    }
+
+    fun allergyEntityListToAllergyList(entities: List<AllergyEntity>): List<Allergy> {
```
rishabh-997, line 274: I did not rename it because, after compilation, generics are erased. So `List<AllergyEntity>` and `List<ObservationEntity>` at **line 71** become just List in the Java bytecode. And we cannot have more than one method with the same signature.

rishabh-997: So either we can add a dummy parameter in this function so that signatures will become different or let it be as it is

openmrs-client/src/main/java/org/openmrs/mobile/api/repository/AllergyRepository.java
```diff
@@ -58,7 +56,7 @@ public AllergyRepository(String id, AllergyRoomDAO allergyRoomDAO) {
     public LiveData<List<Allergy>> getAllergies(RestApi restApi, String uuid) {
         MutableLiveData<List<Allergy>> allergyLiveData = new MutableLiveData<>();
         allergyEntitiesOffline = allergyRoomDAO.getAllAllergiesByPatientID(patientID);
-        allergyLiveData.setValue(allergyEntityListToAllergyList(allergyEntitiesOffline));
+        allergyLiveData.setValue(AppDatabaseHelper.INSTANCE.allergyEntityListToAllergyList(allergyEntitiesOffline));
```
f4ww4z, line 59:
```suggestion
        allergyLiveData.setValue(AppDatabaseHelper.convertTo(allergyEntitiesOffline));
```

openmrs-client/src/main/java/org/openmrs/mobile/databases/AppDatabaseHelper.kt
f4ww4z: In Kotlin this is fine, in Java it's not. Use the `@JvmName` parameter to make the method name different, something like `convertTo` . [See this thread](https://stackoverflow.com/a/51581951/8434188)

rishabh-997:
This `appveyor` is failing a lot since few days...

PR #804, opened 2020-08-13
AC-829 Fixed crash at admission form by using live data

rishabh-997:
We need to observe the encounter roles and locations using live data as in dark mode, activity is re-created immediately and the results cause NPE as the fragment is destroyed on which it needed to show.

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-829

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/api/repository/ProviderRepository.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/api/repository/ProviderRepository.java
@@ -36,8 +36,6 @@ import org.openmrs.mobile.dao.ProviderRoomDAO;
 import org.openmrs.mobile.databases.AppDatabase;
 import org.openmrs.mobile.databases.entities.LocationEntity;
 import org.openmrs.mobile.listeners.retrofitcallbacks.DefaultResponseCallback;
-import org.openmrs.mobile.listeners.retrofitcallbacks.EncounterResponseCallback;
-import org.openmrs.mobile.listeners.retrofitcallbacks.LocationResponseCallback;
 import org.openmrs.mobile.models.Provider;
 import org.openmrs.mobile.models.Resource;
 import org.openmrs.mobile.models.Results;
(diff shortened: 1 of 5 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/api/repository/ProviderRepository.java
```diff
@@ -253,54 +251,50 @@ public void onFailure(@NotNull Call<ResponseBody> call, @NotNull Throwable t) {
...
-                    if (response.isSuccessful()) {
-                        callback.onResponse(response.body().getResults());
-                    } else {
-                        callback.onErrorResponse(OpenMRS.getInstance().getString(R.string.error_occurred));
-                    }
+                if (response.isSuccessful()) {
+                    encounterRoles.setValue(response.body().getResults());
+                } else {
+                    encounterRoles.setValue(null);
                 }
```
LuGO0, line 290: @rishabh-997  how are we taking care of the UI responses to the user after the callback has been removed all the views which implemented this callback wont get any callback to handle ?

rishabh-997: We were using callbacks because we were not observing data. Now that I have changed return type to `LiveData<T>`, we don't need to use callbacks, the UI will simply observe results...
**LiveData** will allow the observer to be removed when the state of the corresponding Lifecycle object changes to **DESTROYED**. This is useful for activities and fragments because they do not need to worry about leaks—activities and fragments are instantly unsubscribed when their lifecycles are destroyed.

LuGO0: yes @rishabh-997 its good actually bt wont it require some changes in the code calling these methods on fragments and activities? since they must be still have implemented these callbacks i coudnt see changes done in presenter arent they requered.
am I getting something wrong here ?

LuGO0: yes @rishabh-997  actually i missed the changes you made in presenter I thought they were missing .

PR #817, opened 2020-10-02
AC-792 Fixing Back Button Intent for Contact Us Activity

Bisht13:

Added intent to the back button of the action bar in Contact Us activity.

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-792

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/community/contact/ContactUsActivity.kt
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/community/contact/ContactUsActivity.kt
@@ -21,6 +21,7 @@ import android.view.Menu
 import android.view.MenuItem
 import org.openmrs.mobile.R
 import org.openmrs.mobile.activities.ACBaseActivity
+import org.openmrs.mobile.activities.login.LoginActivity
 import org.openmrs.mobile.databinding.ActvityContactUsBinding
 import org.openmrs.mobile.utilities.ToastUtil
 
@@ -79,6 +80,8 @@ class ContactUsActivity : ACBaseActivity(), ContactUsContract.View {
 
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         if (item.itemId == android.R.id.home) {
+            val backButtonIntent = Intent(this, LoginActivity::class.java)
+            startActivity(backButtonIntent)
             finish()
             return true
         }


openmrs-client/src/main/java/org/openmrs/mobile/activities/community/contact/ContactUsActivity.kt
```diff
@@ -79,6 +80,8 @@ class ContactUsActivity : ACBaseActivity(), ContactUsContract.View {
 
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         if (item.itemId == android.R.id.home) {
+            val backButtonIntent = Intent(this, LoginActivity::class.java)
```
rishabh-997, line 83: This seems to be a work around... You need to find the stack trace to get to the reason behind this behaviour.... What you did have following mistakes...
1. no matter at which activity the user is, pressing back icon will always bring him to login screen
2. You have just added intent on menu button, what if the user clicks back button from bottom? your work around will fail...

rishabh-997 (changes requested):
Try to find the reason behind this... why is previous activity finished without destroying it

openmrs-client/src/main/java/org/openmrs/mobile/activities/community/contact/ContactUsActivity.kt
Bisht13: I tried finding why the saved instance of login activity is not being called. The activity is there in the stack when the contact activity is created but is popped from the stack when the back button is clicked. I was unable to recreate your first issue. The back button of other action bars don't call back to the login page. The second issue called be fixed by using the same work around in onBackPressed function in contact class.

rishabh-997: > doesn't call back to the login page

That is because you are logged in and redirected to the dashboard. Consider an instance when you are on the visits page and go to contacts from there. Pressing back will take you to the dashboard rather than the visits page. Try to correct it the right way instead of using workarounds. :100:

rishabh-997:
Any updates on this?

Bisht13:
I'm having my college counselling this month, so I might not be able to work on it for some time.

rishabh-997:
> I'm having my college counselling this month

That's great, fill your choices carefully, and welcome to college life :partying_face:

rishabh-997:
Did you happen to find some fix for this?

rishabh-997:
Closing due to inactivity. Also, the issue is fixed at #822

PR #820, opened 2020-11-13
AC-851: Fix errors in inflating attr when using on View's background.

manjotsidhu:

When using attribute as a View's background, the app crashes failing to inflate the View

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-851

### DIFF
--- a/openmrs-client/src/main/res/drawable-hdpi/rectangle_shape_with_primary_text_color.xml
+++ b/openmrs-client/src/main/res/drawable-hdpi/rectangle_shape_with_primary_text_color.xml
@@ -1,0 +1,3 @@
+<shape xmlns:android="http://schemas.android.com/apk/res/android" android:shape="rectangle">
+    <solid android:color="?attr/textColorPrimary" />
+</shape>
--- a/openmrs-client/src/main/res/layout/activity_patient_dashboard.xml
+++ b/openmrs-client/src/main/res/layout/activity_patient_dashboard.xml
@@ -49,7 +49,7 @@
             <View
                 android:layout_width="match_parent"
                 android:layout_height="1dp"
-                android:background="?android:attr/textColorPrimary" />
+                android:background="@drawable/rectangle_shape_with_primary_text_color" />
 
             <androidx.viewpager.widget.ViewPager
                 android:id="@+id/pager"
--- a/openmrs-client/src/main/res/layout/fragment_patient_details.xml
+++ b/openmrs-client/src/main/res/layout/fragment_patient_details.xml
@@ -122,7 +122,7 @@
             android:layout_height="1dp"
             android:layout_marginHorizontal="@dimen/padding_16"
             android:layout_marginTop="24dp"
-            android:background="?android:attr/textColorPrimary"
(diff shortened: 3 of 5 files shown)


LuGO0:
@manjotsidhu while doing this fix what problem you came across, what was the Issue with the earlier attributes? please explain your change and link this PR to the JIRA issue.

manjotsidhu:
> @manjotsidhu while doing this fix what problem you came across, what was the Issue with the earlier attributes? please explain your change and link this PR to the JIRA issue.

App was getting crashed when loading Patients Dashboard activity faced by me and @LuGO0 . The logs are available at JIRA issue. This issue was due to the color attribute being applied to View's background. 

This thread also explains the same issue : https://stackoverflow.com/questions/17843635/how-to-use-textcolorprimary-as-background-color-in-a-style

LuGO0:
@rishabh-997 The fix works for me but I dont know if there is some more elegant way of doing this rather than creating a whole new xml file ?

manjotsidhu:
> @rishabh-997 The fix works for me but I dont know if there is some more elegant way of doing this rather than creating a whole new xml file ?

The recommended background to any View should be a drawable and the only way to define a drawable is to create a drawable with that color. 

I would give more information on this, API <= 21 doesn't allows referencing a platform attribute in style. Therefore, for backward compatibility we have to explicitly define a reference to that variable in `attrs.xml`.

LuGO0:
@manjotsidhu @rishabh-997 is [this](https://github.com/LuGO0/openmrs-contrib-android-client/commit/bee0f9468e25b58da606357f9b3f0fb4da212e2a) a good way to fix this bug ??
I really feel the current fix is an overkill, @manjotsidhu  if the fix looks good and is approved by @rishabh-997 you can cherry pick the commits.

manjotsidhu:
> @manjotsidhu @rishabh-997 is [this](https://github.com/LuGO0/openmrs-contrib-android-client/commit/bee0f9468e25b58da606357f9b3f0fb4da212e2a) a good way to fix this bug ??
> I really feel the current fix is an overkill, @manjotsidhu  if the fix looks good and is approved by @rishabh-997 you can cherry pick the commits. 

Not sure, the attribute which was created I guess is for dynamic theme changes that's why I think we can't use a static color instead of the attribute.

LuGO0:
> Not sure, the attribute which was created I guess is for dynamic theme changes that's why I think we can't use a static color instead of the attribute.

I looked up the styles theme and I found 2 commented lines which contain these attributes, I dont know why are they commented maybe they would be satisfy the need.
here https://github.com/openmrs/openmrs-contrib-android-client/blob/master/openmrs-client/src/main/res/values/styles.xml#L31

lovingroshni15 (approved):
LGTM !!! Was quite busy with university exams so this review got a bit late :sweat_smile:

rishabh-997 (approved):
LGTM !!! Was quite busy with university exams so this review got a bit late :sweat_smile: 
I guess I was logged in with the wrong ID previously.

PR #854, opened 2021-03-01
AC-886 using viewBinding for ProviderManagerDashboard package

LuGO0:
1. Used view Binding in both Activity as well as Fragment.

## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-886

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/providermanagerdashboard/ProviderManagerDashboardFragment.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/providermanagerdashboard/ProviderManagerDashboardFragment.java
@@ -30,9 +30,9 @@ import androidx.swiperefreshlayout.widget.SwipeRefreshLayout;
 
 import com.google.android.material.floatingactionbutton.FloatingActionButton;
 
-import org.openmrs.mobile.R;
 import org.openmrs.mobile.activities.ACBaseFragment;
 import org.openmrs.mobile.activities.providermanagerdashboard.addprovider.AddProviderActivity;
+import org.openmrs.mobile.databinding.FragmentProviderManagementBinding;
 import org.openmrs.mobile.models.Provider;
 import org.openmrs.mobile.utilities.ApplicationConstants;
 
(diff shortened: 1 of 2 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/activities/providermanagerdashboard/ProviderManagerDashboardFragment.java
```diff
@@ -47,11 +47,11 @@
     implements ProviderManagerDashboardContract.View {
 
     // Fragment components
+    private FragmentProviderManagementBinding binding;
```
amtheshubham, line 50: @LuGO0 I think you should make the `binding` variable nullable and set it **`null`** in Fragment's **`onDestroyView()`** as not doing so will result in **_memory leak_** because Fragments outlive their views.

Now that I have spotted this, I will create an issue also for the fragments where the **ViewBinding** is used but never been set to `null` in **`onDestroyView()`**

LuGO0: Good catch @amtheshubham but I think there is a more elegant way of handling these memory leaks and that is moving the binding reference from fragment scope to function scope. I wouldnt want to unnecessariliy override the onDestroyView() in every fragment.
https://dev.to/s_anastasov/fragments-viewbinding-and-memory-leaks-3kn1

amtheshubham: This is efficient but I guess that's meant for _MVVM_ pattern as stated in the article itself. We have more than one functions which are using binding in the fragment. Though, for this particular `ProviderManagerDashboardFragment` you can follow this _function scope_ approach. 
so according to it, please move the declaration of binding to `onCreateView()` of the Fragment

LuGO0: correct

amtheshubham (approved):
LGTM!

openmrs-client/src/main/java/org/openmrs/mobile/activities/providermanagerdashboard/ProviderManagerDashboardFragment.java
rishabh-997: do it and rest LGTM, good catch @amtheshubham

LuGO0:
Yes @rishabh-997 it works as expected.

openmrs-client/src/main/java/org/openmrs/mobile/activities/providermanagerdashboard/ProviderManagerDashboardFragment.java
LuGO0: its done @rishabh-997

PR #883, opened 2021-03-25
AC-946: Solved the bug in language switching feature.

Akshatji800:
## Issue I worked on

JIRA Issue: https://issues.openmrs.org/browse/AC-946

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/activities/settings/SettingsPresenter.kt
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/activities/settings/SettingsPresenter.kt
@@ -81,11 +81,9 @@ class SettingsPresenter(private val mSettingsView: SettingsContract.View, privat
 
     override val languagePosition: Int
         get() {
-            val lang = LanguageUtils.getLanguage()
-            val languageList = ApplicationConstants.OpenMRSlanguage.LANGUAGE_LIST
             var i = 0
-            while (i < languageList.size) {
-                if (lang == languageList[i]) {
+            while (i < ApplicationConstants.OpenMRSlanguage.LANGUAGE_CODE.size) {
+                if (LanguageUtils.getLanguage()== ApplicationConstants.OpenMRSlanguage.LANGUAGE_CODE[i]) {
                     return i
                 }
                 i++


Akshatji800:
I know the method I have used to solve the issue is a bad coding practice but still, it solves the issue without creating much changes in the code base and is the only method I found till now to solve the issue without unnecessary changes in code. Suggestions to have a better solution are welcomed.
For more information on this issue and to view the GIF to understand the issue and to view the status of language changing feature through a GIF after the changes committed are made one can visit the issue link: [here](https://issues.openmrs.org/browse/AC-946). The GIFs are attached in the attachment section.

openmrs-client/src/main/java/org/openmrs/mobile/activities/settings/SettingsPresenter.kt
```diff
@@ -84,13 +84,14 @@ class SettingsPresenter(private val mSettingsView: SettingsContract.View, privat
             val lang = LanguageUtils.getLanguage()
             val languageList = ApplicationConstants.OpenMRSlanguage.LANGUAGE_LIST
             var i = 0
-            while (i < languageList.size) {
-                if (lang == languageList[i]) {
-                    return i
-                }
-                i++
+            if (lang == "hi")
```
HerbertYiga, line 87: @Akshatji800  i think we need some work around on this so that we dont hard code things

Akshatji800: @HerbertYiga gave a better solution without hard coding things you can have a check now.

openmrs-client/src/main/java/org/openmrs/mobile/activities/settings/SettingsPresenter.kt
```diff
@@ -78,14 +78,15 @@ class SettingsPresenter(private val mSettingsView: SettingsContract.View, privat
         set(lang) {
             LanguageUtils.setLanguage(lang)
         }
+    val languageSetArray = arrayOf<String>("en","hi")
```
LuGO0, line 82: thats pretty much still hard coded

Akshatji800: @LuGO0 can you now please review again is it still hard-coded or not I have tried to give a better solution.

LuGO0:
@Akshatji800 why was this Issue happening in the first place could you link the PR to the original change which might have cause this Issue? would be easier to review that way. Thank you !!

Akshatji800:
@rishabh-997 can you please review.

Akshatji800:
@LuGO0 this is also important and solves a bug in the app. Can you review this also please.

Akshatji800:
> @Akshatji800 why was this Issue happening in the first place could you link the PR to the original change which might have cause this Issue? would be easier to review that way. Thank you !!

In this [PR](https://github.com/rishabh-997/openmrs-contrib-android-client/commit/f71dc53d0004d765957506ba235986f90265d684) the language codes were changed to full names to display better in the language change spinner button but it somehow the fact that it should also be changed while comparing is missed so it had caused the bug in my opinion.

LuGO0:
I think the app works fine  without this change if not please get this Issue clarified in the JIRA.

LuGO0:
I checked on the device works fine !!

PR #891, opened 2021-04-10
AC-968/AC-969/AC-970: Fix Provider module offline support issues

amrsalah3:

First, these 3 issues are related to each other and the same classes (and lines of code) need to be changed, so I've combined them in one PR for not producing merge conflicts.
1. Fixed provider's name being _[givenName givenName]_ instead of _[givenName familyName]_ in server when reconnecting to the internet. [AC-968]
2. Fixed AddProviderWorker being called forever even after success. [AC-969]
3. Fixed failing to edit or delete a newly created Provider in both online and offline modes. [AC-970]

## Issue I worked on

JIRA Issue:
https://issues.openmrs.org/browse/AC-968
https://issues.openmrs.org/browse/AC-969
https://issues.openmrs.org/browse/AC-970

### DIFF
--- a/openmrs-client/src/main/java/org/openmrs/mobile/api/repository/ProviderRepository.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/api/repository/ProviderRepository.java
@@ -116,11 +116,7 @@ public class ProviderRepository extends BaseRepository {
                 public void onResponse(@NotNull Call<Provider> call, @NotNull Response<Provider> response) {
                     if (response.isSuccessful()) {
                         //offline adding provider
-                        provider.setUuid(response.body().getUuid());
-                        provider.setId(providerRoomDao.addProvider(provider));
-                        //editing the provider
-                        providerRoomDao.updateProviderByUuid(response.body().getDisplay(), provider.getId(), response.body().getPerson(), response.body().getUuid(),
-                            response.body().getIdentifier());
+                        providerRoomDao.addProvider(response.body());
 
                         ToastUtil.success(openMrs.getString(R.string.add_provider_success_msg));
                         logger.e("Adding Provider Successful " + response.raw());
@@ -140,10 +136,7 @@ public class ProviderRepository extends BaseRepository {
             long providerId = providerRoomDao.addProvider(provider);
 
             //delegate to the workManager
-            Data data = new Data.Builder().putString("first_name", provider.getPerson().getName().getGivenName())
-                .putString("last_name", provider.getPerson().getName().getFamilyName())
-                .putString("identifier", provider.getIdentifier())
-                .putLong("id", providerId).build();
+            Data data = new Data.Builder().putLong("id", providerId).build();
 
             callback.onResponse();
 
--- a/openmrs-client/src/main/java/org/openmrs/mobile/api/repository/ProviderRepository.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/api/repository/ProviderRepository.java
@@ -116,8 +116,8 @@ public class ProviderRepository extends BaseRepository {
                 public void onResponse(@NotNull Call<Provider> call, @NotNull Response<Provider> response) {
                     if (response.isSuccessful()) {
                         //offline adding provider
+                        provider.setUuid(response.body().getUuid());
                         provider.setId(providerRoomDao.addProvider(provider));
-
                         //editing the provider
                         providerRoomDao.updateProviderByUuid(response.body().getDisplay(), provider.getId(), response.body().getPerson(), response.body().getUuid(),
                             response.body().getIdentifier());
@@ -137,11 +137,11 @@ public class ProviderRepository extends BaseRepository {
         } else {
 
             //offline addition operation
-            Long providerId = providerRoomDao.addProvider(provider);
+            long providerId = providerRoomDao.addProvider(provider);
 
             //delegate to the workManager
             Data data = new Data.Builder().putString("first_name", provider.getPerson().getName().getGivenName())
-                .putString("last_name", provider.getPerson().getName().getGivenName())
+                .putString("last_name", provider.getPerson().getName().getFamilyName())
                 .putString("identifier", provider.getIdentifier())
                 .putLong("id", providerId).build();
 
--- a/openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/AddProviderWorker.java
+++ b/openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/AddProviderWorker.java
@@ -15,14 +15,14 @@
 package org.openmrs.mobile.api.workers.provider;
 
 import android.content.Context;
+import android.os.Handler;
+import android.os.Looper;
 
 import androidx.annotation.NonNull;
 import androidx.work.Worker;
 import androidx.work.WorkerParameters;
 
-import org.jetbrains.annotations.NotNull;
 import org.openmrs.mobile.R;
-import org.openmrs.mobile.listeners.retrofitcallbacks.CustomResponseCallback;
 import org.openmrs.mobile.api.RestApi;
 import org.openmrs.mobile.api.RestServiceBuilder;
 import org.openmrs.mobile.application.OpenMRS;
@@ -34,11 +34,10 @@ import org.openmrs.mobile.models.Provider;
 import org.openmrs.mobile.utilities.NetworkUtils;
 import org.openmrs.mobile.utilities.ToastUtil;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
-import retrofit2.Call;
-import retrofit2.Callback;
 import retrofit2.Response;
 
 public class AddProviderWorker extends Worker {
@@ -54,28 +53,20 @@ public class AddProviderWorker extends Worker {
     @NonNull
     @Override
     public Result doWork() {
-        final boolean[] result = new boolean[1];
         String firstName = getInputData().getString("first_name");
         String lastName = getInputData().getString("last_name");
         String identifier = getInputData().getString("identifier");
         Person person = createPerson(firstName, lastName);
         Provider providerTobeCreated = createNewProvider(person, identifier, getInputData().getLong("id", 0l));
 
-        addProvider(restApi, providerTobeCreated, new CustomResponseCallback() {
-            @Override
-            public void onResponse() {
-                result[0] = true;
+        boolean result = addProvider(restApi, providerTobeCreated);
+
+        if (result) {
+            new Handler(Looper.getMainLooper()).post(() -> {
                 ToastUtil.success(OpenMRS.getInstance().getString(R.string.add_provider_success_msg));
                 OpenMRS.getInstance().getOpenMRSLogger().e("Adding Provider Successful ");
-            }
+            });
 
-            @Override
-            public void onErrorResponse() {
-                result[0] = false;
-            }
-        });
-
-        if (result[0]) {
             return Result.success();
         } else {
             return Result.retry();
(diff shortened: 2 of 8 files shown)


openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/AddProviderWorker.java
```diff
@@ -34,101 +30,95 @@
...
+			try {
+				// Execute synchronous API call to block Worker's return statement until done.
+				Response<Provider> response = restApi.addProvider(provider).execute();
+				if (response.isSuccessful()) {
+					//offline updating the uuid and properties of the provider
+					providerRoomDao.updateProviderUuidById(provider.getId(), response.body().getUuid());
+					providerRoomDao.updateProviderByUuid(response.body().getDisplay(), provider.getId(),
+					    response.body().getPerson(), response.body().getUuid(), response.body().getIdentifier());
+					return true;
+				}
```
LuGO0, line 116: @SalahoAmro should we use either updateByUUID or updateById like we are updating uuid by id and then provider by uuid, can we simply update provider by id? will there be any Issues?

LuGO0:
looks good to me I wanted to know more about the Id issue how did you get over it?

openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/AddProviderWorker.java
```diff
@@ -34,101 +30,95 @@
...
-                }
-
-                @Override
-                public void onFailure(@NotNull Call<Provider> call, @NotNull Throwable t) {
-                    callback.onErrorResponse();
-                }
-            });
-        }
-    }
+
```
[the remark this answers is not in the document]
SalahoAmro: Done!

openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/AddProviderWorker.java
SalahoAmro: The method **updateProviderById()** searches for Provider by (_id_) field of Provider object, which is always empty always until we update it because the primary key of the provider is (__id_) field of its superclass _Resource_.
When we create a Provider object in AddProviderWorker using the data passed to the worker, the passed id is actually the __id_ NOT _id_. You can notice this in ProviderRepository in lines 140 and 146. So until now, _id_ field of this provider in the database table is empty. So we first, **updateProviderUuidById()** when connecting to the internet, then update rest of the data of the provider in the database using updateProviderById() which also updates the empty _id_ of the provider (not useful anyway because we already have __id_ value).

This is confusing, to be honest. IMO, Provider DAO methods should all be updated to use either _id or id, but first, we need to know what is actually the difference between them because _id is the primary key so we have to use it now.

SalahoAmro:
@LuGO0 I replied to your comment in the review about this. Also, I want to add that I tried running current tests in ProviderRoomDAOTest and they fail although Travis passes these tests! I mentioned that earlier here in Talk: https://talk.openmrs.org/t/gsoc-2021-openmrs-android-client-project-3-0-x-a-problem-with-testing/32786. So we can raise an issue if you kindly agree with me. Also, as I mentioned _id and id fields are really confusing and may be responsible for these tests failure, I tried my best to understand them and this PR is the only way to solve these issues without refactoring current Provider DAO methods that use _id_ field in all queries.

openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/AddProviderWorker.java
LuGO0: can we probe into this @SalahoAmro what are other entities using in this. I think we should draw a conclusion here and fix this problem, I am also confused by this for now.

LuGO0: This might help @SalahoAmro https://github.com/openmrs/openmrs-core/tree/master/api/src/main/java/org/openmrs

SalahoAmro: @LuGO0 I have already read the core data model and REST docs, _id is used in the server. We can notice this when fetching providers from the server for the first time, I inspected Android Client's database and found that _id fields are updated, whereas, id fields are empty.

LuGO0:
I dont think we should merge this without understanding, will it be a huge change for us to start using _id instead of id because as I see in the openMRS models _id is used and id makes little sense to me atleast. sorry for the late reply

SalahoAmro:
> I dont think we should merge this without understanding, will it be a huge change for us to start using _id instead of id because as I see in the openMRS models _id is used and id makes little sense to me atleast. sorry for the late reply :(

I agree with you. We need to understand that first, but I wonder why doesn't anyone reply to us? :(

LuGO0:
Hi @amrsalah3 were you able to figure out if there were really multiple providers added or just multiple API calls intercepted in chucker, i.e. was there an Issue with the chucker or something to do with offline functionality?

amrsalah3:
@LuGO0 It was some problems with the offline structure, not chucker.

openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/AddProviderWorker.java
```diff
@@ -54,28 +53,20 @@ public AddProviderWorker(@NonNull Context context, @NonNull WorkerParameters wor
...
 
-        addProvider(restApi, providerTobeCreated, new CustomResponseCallback() {
-            @Override
-            public void onResponse() {
-                result[0] = true;
+        boolean result = addProvider(restApi, providerTobeCreated);
+
+        if (result) {
+            new Handler(Looper.getMainLooper()).post(() -> {
                 ToastUtil.success(OpenMRS.getInstance().getString(R.string.add_provider_success_msg));
```
LuGO0, line 66: Is it the Fix for the Issue we were facing @amrsalah3 the infinite calls?

amrsalah3: Yes, asynchronous calls of the API were preventing the Success result from being returned from the the worker so it was retrying forever.
Now, I've replaced it with a synchronous call.

amrsalah3: @LuGO0 check addProvider method, too. API call is modified inside it.

LuGO0: perfect I was just trying this branch out locally.

openmrs-client/src/main/java/org/openmrs/mobile/dao/ProviderRoomDAO.java
```diff
@@ -58,6 +58,9 @@
     @Query("UPDATE provider_table SET person= :person, id= :id, display=:display, identifier=:identifier WHERE uuid = :uuid")
     void updateProviderByUuid(String display, long id, Person person, String uuid, String identifier);
 
+    @Query("UPDATE provider_table SET uuid=:uuid WHERE _id=:id")
+    void updateProviderUuidById(long id, String uuid);
+
     @Query("SELECT * FROM provider_table")
```
LuGO0, line 64: I will work a bit more here and experiment out a bit with _id and id Issue before

amrsalah3: Thank you!

LuGO0:
Hi @amrsalah3 made some changes please review 😁

openmrs-client/src/main/java/org/openmrs/mobile/api/repository/ProviderRepository.java
```diff
@@ -116,11 +116,7 @@ public void addProvider(Provider provider, DefaultResponseCallback callback) {
                 public void onResponse(@NotNull Call<Provider> call, @NotNull Response<Provider> response) {
                     if (response.isSuccessful()) {
                         //offline adding provider
-                        provider.setId(providerRoomDao.addProvider(provider));
-
-                        //editing the provider
-                        providerRoomDao.updateProviderByUuid(response.body().getDisplay(), provider.getId(), response.body().getPerson(), response.body().getUuid(),
-                            response.body().getIdentifier());
+                        response.body().setId(providerRoomDao.addProvider(response.body()));
```
LuGO0, line 120: I think this is wrong here the ID wont get setup in the db since its only for the response object

openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/AddProviderWorker.java
```diff
@@ -54,81 +49,77 @@ public AddProviderWorker(@NonNull Context context, @NonNull WorkerParameters wor
...
+//        Provider provider = new Provider();
+//        provider.setPerson(person);
+//        provider.setUuid(null);
+//        provider.setIdentifier(identifier);
+//        provider.setRetired(false);
+//        provider.setId(providerId);
+//
+//        return provider;
+//    }
+
```
LuGO0, line 106: will do cleanup later.

openmrs-client/src/main/java/org/openmrs/mobile/api/repository/ProviderRepository.java
```diff
@@ -116,11 +116,7 @@ public void addProvider(Provider provider, DefaultResponseCallback callback) {
                 public void onResponse(@NotNull Call<Provider> call, @NotNull Response<Provider> response) {
                     if (response.isSuccessful()) {
                         //offline adding provider
-                        provider.setUuid(response.body().getUuid());
-                        provider.setId(providerRoomDao.addProvider(provider));
-                        //editing the provider
-                        providerRoomDao.updateProviderByUuid(response.body().getDisplay(), provider.getId(), response.body().getPerson(), response.body().getUuid(),
-                            response.body().getIdentifier());
+                        response.body().setId(providerRoomDao.addProvider(response.body()));
```
amrsalah3: What I understand now that there will be an auto-generated primary key __id_ and you deleted the obsolete _id_, so why you set the generated id to the response provider's object instead of just adding provider to Room?

openmrs-client/src/main/java/org/openmrs/mobile/dao/ProviderRoomDAO.java
```diff
@@ -55,7 +55,7 @@
      */
 
     @TypeConverters(PersonConverter.class)
-    @Query("UPDATE provider_table SET person= :person, id= :id, display=:display, identifier=:identifier WHERE uuid = :uuid")
+    @Query("UPDATE provider_table SET person= :person, _id= :id, display=:display, identifier=:identifier WHERE uuid = :uuid")
     void updateProviderByUuid(String display, long id, Person person, String uuid, String identifier);
 
     @Query("UPDATE provider_table SET uuid=:uuid WHERE _id=:id")
```
amrsalah3, line 61: I think this method that I created is no longer needed.`updateProviderUuidById()`

amrsalah3:
@LuGO0 Perfect work! :smiley:

LuGO0:
Hi @rishabh-997 @amrsalah3 all 3 operations work fine for me now. could you please check once I have squashed all commits into 2 commits.

openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/AddProviderWorker.java
```diff
@@ -54,81 +49,41 @@ public AddProviderWorker(@NonNull Context context, @NonNull WorkerParameters wor
...
-            }
-        });
-
-        if (result[0]) {
-            return Result.success();
+        if (provider == null) {
+            return Result.failure();
         } else {
-            return Result.retry();
+            // Its necessary since server gives Error 500 if request has UUID set as ""
```
rishabh-997, line 57: Can you remove this comment

openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/AddProviderWorker.java
```diff
@@ -54,81 +49,41 @@ public AddProviderWorker(@NonNull Context context, @NonNull WorkerParameters wor
...
+            provider.getPerson().setUuid(null);
+
+            if (addProvider(restApi, provider)) {
+                new Handler(Looper.getMainLooper()).post(() -> {
+                    ToastUtil.success(OpenMRS.getInstance().getString(R.string.add_provider_success_msg));
+                    OpenMRS.getInstance().getOpenMRSLogger().e("Adding Provider Successful ");
```
rishabh-997, line 63: Use string resource file

openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/UpdateProviderWorker.java
```diff
@@ -49,53 +49,40 @@ public UpdateProviderWorker(@NonNull Context context, @NonNull WorkerParameters
...
-        });
+        Provider provider = providerRoomDao.findProviderByUUID(providerUuidTobeUpdated).blockingGet();
 
-        if (result[0]) {
-            return Result.success();
+        if (provider == null) {
+            return Result.failure();
         } else {
-            return Result.retry();
+            // Its necessary since server gives Error 500 if request has UUID set as ""
```
rishabh-997, line 58: Again, remove this

openmrs-client/src/main/java/org/openmrs/mobile/api/workers/provider/UpdateProviderWorker.java
```diff
@@ -49,53 +49,40 @@ public UpdateProviderWorker(@NonNull Context context, @NonNull WorkerParameters
...
+            provider.getPerson().setUuid(null);
+
+            if (updateProvider(restApi, provider)) {
+                new Handler(Looper.getMainLooper()).post(() -> {
+                    ToastUtil.success(OpenMRS.getInstance().getString(R.string.edit_provider_success_msg));
+                    OpenMRS.getInstance().getOpenMRSLogger().e("Editing Provider Successful ");
```
rishabh-997, line 64: String resource file

openmrs-client/src/main/java/org/openmrs/mobile/models/Provider.kt
```diff
@@ -16,7 +16,6 @@ package org.openmrs.mobile.models
 
 import androidx.room.ColumnInfo
 import androidx.room.Entity
-import androidx.room.PrimaryKey
 import androidx.room.TypeConverters
 import com.google.gson.annotations.Expose
 import com.google.gson.annotations.SerializedName
```
rishabh-997, line 21: Can you remove star imports

rishabh-997 (changes requested):
Rest LGTM. I'll just check the changes locally once

LuGO0:
Yes @rishabh-997 I will make these changes please check once locally.

LuGO0:
@rishabh-997 made required changes please review

rishabh-997 (approved):
LGTM... thanks for the efforts @LuGO0 @amrsalah3

PR #894, opened 2021-05-07
AC-948 Migrated Logs Package from MVP to MVVM

shivamsawlani02:
Migrated Logs Package from MVP to MVVM. 
Removed the Contract and Presenter Classes and added a ViewModel class.

## Issue I worked on
JIRA Issue: https://issues.openmrs.org/browse/AC-948


shivamsawlani02:
@LuGO0 @rishabh-997 please review.

openmrs-client/src/main/java/org/openmrs/mobile/activities/logs/LogsFragment.kt
```diff
@@ -21,31 +21,47 @@ import android.view.LayoutInflater
...
+    private fun attachLogsToTextView() {
+        mViewModel.getLogs()
+        mViewModel.logsText.observe(viewLifecycleOwner, Observer { updatedLogTexts ->
+            binding.tvLogs.text = updatedLogTexts
+        })
+    }
+
+    private fun fabCopyAll() {
+        mViewModel.getLogs()
+        mViewModel.logsText.observe(viewLifecycleOwner, Observer { updatedLogTexts ->
```
HerbertYiga, line 60: thanks @LuGO0  how do things look to you on this entire pr?

LuGO0:
@shivamsawlani02 we need to start from the bottoms up I think making a base class structure before adding in classes on top of it. I think this wont work.

shivamsawlani02:
> @shivamsawlani02 we need to start from the bottoms up I think making a base class structure before adding in classes on top of it. I think this wont work.

Okay, @LuGO0. I will create a new issue in JIRA for creating Base classes for Fragments and ViewModels. There isn't any change in Activities, so I think we don't need a new BaseActivity. So, will creating a new BaseFragment and a BaseViewModel be sufficient, or do we need to do something else also?

LuGO0:
yes I think so then I will have to keep this Issue on hold uptill then since it will be based on those base classes rather than be independent, and you might want to have a look we have a BaseFragment as well if that helps.

shivamsawlani02:
> yes I think so then I will have to keep this Issue on hold uptill then since it will be based on those base classes rather than be independent, and you might want to have a look we have a BaseFragment as well if that helps.

Yes, we can keep this on hold till base classes are created.

LuGO0:
closing this for now.

